Issue #8, opened 2015-07-10
Using UFIXIT when there are multiple of the same type of error on the page adds "Fixed" to all of the errors.


bagofarms:
<img width="351" alt="screenshot 2015-07-10 14 42 20" src="https://cloud.githubusercontent.com/assets/1225334/8626593/6ea744e8-2713-11e5-8f42-b56ca65bca30.png">

techrangerfj:
Resolution merged to master. See commit #0a2302fce225bccdc1635979c59546eba86da727

techrangerfj:
Resolution merged to master. See commit #0a2302fce225bccdc1635979c59546eba86da727

PR #21, opened 2015-08-19
Issue #8 Resolution

techrangerfj:
Fixing one instance of multiple for a particular error on a particular page now no longer updates all instances of the error, but rather the one that has just been fixed. Link to 'Close view preview' no added above preview. Updates #8


bagofarms:
The description of the error disappears after the error is fixed.  Make it so that the description stays visible after the error has been fixed.

Issue #9, opened 2015-07-10
"No table headers found" in a table that already has headers (see comments for sample code)


bagofarms:
<table style="width: 551px; margin-left: auto; margin-right: auto;" border="1" cellspacing="0" cellpadding="0">
<tbody>
<tr>
<th style="text-align: center;" width="84">
<p>Week/Dates</p>
</th>
<th width="53">
<p>Mode</p>
</th>
<th width="113">
<p>Topic</p>
</th>
<th width="144">
<p>Assignments</p>
</th>
<th width="157">
<p>Due Dates</p>
</th>
</tr>
<tr>
<td width="84">
<p>Week 1</p>
</td>
<td valign="top" width="53">
<p>F2F</p>
</td>
<td valign="top" width="113">
<p>&nbsp;Course Orientation</p>
</td>
<td valign="top" width="144">
<p>Syllabus Quiz</p>
</td>
<td valign="top" width="157">
<p>&nbsp;<span>8/31 at 11:55pm EST</span></p>
</td>
</tr>
<tr>
<td width="84">
<p>Week 2</p>
</td>
<td valign="top" width="53">
<p>Online</p>
</td>
<td valign="top" width="113">
<p>&nbsp;</p>
</td>
<td valign="top" width="144">
<p>&nbsp;</p>
</td>
<td valign="top" width="157">
<p>&nbsp;</p>
</td>
</tr>
<tr>
<td width="84">
<p>Week 3</p>
<p>&nbsp;</p>
</td>
<td valign="top" width="53">
<p>F2F</p>
</td>
<td valign="top" width="113">
<p>&nbsp;</p>
</td>
<td valign="top" width="144">
<p>&nbsp;</p>
</td>
<td width="157">
<p>&nbsp;</p>
</td>
</tr>
<tr>
<td width="84">
<p>Week 4</p>
<p>&nbsp;</p>
</td>
<td valign="top" width="53">
<p>Online</p>
</td>
<td valign="top" width="113">
<p>&nbsp;</p>
</td>
<td valign="top" width="144">
<p>&nbsp;</p>
</td>
<td width="157">
<p>&nbsp;</p>
</td>
</tr>
<tr>
<td width="84">
<p>Week 5</p>
<p>&nbsp;</p>
</td>
<td valign="top" width="53">
<p>F2F</p>
</td>
<td valign="top" width="113">
<p>&nbsp;</p>
</td>
<td valign="top" width="144">
<p>&nbsp;</p>
[...]

PR #22, opened 2015-08-19
Issue #9 Resolution

techrangerfj:
Verbiage changed for test 'tableThShouldHaveScope' to make it clear that the error is triggered by the non-existence of scope declarations. Updates #9


bagofarms:
The extra quotes are messing up the data structure of en.txt.  It displays with an extra quote on the line after the error description:  https://www.dropbox.com/s/t04wxvasefkxovs/Screenshot%202015-08-19%2016.00.20.png?dl=0

Issue #28, opened 2015-08-28
Only the first image is fixed using UFIXIT


bagofarms:
On a page with multiple images with alt text errors, UFIXIT will work on the first one, but all subsequent images will not have any changes made, even though UFIXIT reports "Fixed"

techrangerfj:
Resolution to issue merged

PR #30, opened 2015-09-08
Updates Issue #28

techrangerfj:
See commit messages


bagofarms:
If you reload the content page after each UFIXIT is run, it works fine.  But, if you do all three UFIXITs first, then reload the content page, all of the alt attributes are still empty.

Issue #20, opened 2015-08-13
Incorrect filename for PDF when running two scans at once.


bagofarms:
To reproduce:
1. Open course A and start a scan, then open course B and start a scan.
2. Course A finishes first, so you click the "Save report as PDF" button.
3. The PDF will download with the filename being based on course B rather than course A.

techrangerfj:
Solution merged in Merge Request 32

PR #32, opened 2015-09-22
Fixes Issue #20

techrangerfj:


bagofarms:
Please revert the user_id to be pulled from the Session rather than storing it in the HTML.

techrangerfj:
@bagofarms user_id is no longer outward facing

PR #44, opened 2015-10-23
use bower for javascript, remove unused jquery ad jspdf libs

iturgeon:
Local jquery lib was being included but not used.  Same with jspdf.

This moves the only remaining js library jscolor out into a bower installed component.  Bower is the de facto js lib manager, so I went with that (but it will require node and bower to use).

I'll add comments to the readme if you want to merge this in (about the bower/node requirements)

This also updates the composer.json to add more project info and hook bower into the install/update command.


bagofarms:
Tested and ready to go, but i'd like you to add the bower/node requirements language you mentioned to the README before i merge.

iturgeon:
Should be pretty close now.

PR #46, opened 2015-10-23
Use a templating engine to clean up the main script logic

iturgeon:
Adds Plates for simple templating
Re-orders includes/requires
Switches include_once to require_once
Changes if else flow at the top to simplify
Adds PDO extension as a composer requirement


bagofarms:
Please merge the current version of Master into your branch and resolve any conflicts

iturgeon:
Updated, looks, conflict was due to the referrer changes, I'm pretty sure I fixed them all up

bagofarms:
Looks like there's still a conflict.  Not sure if that was caused by the "use bower for javascript" pull request i just completed.

bagofarms:
Line 12 of templates/udoit.php references "$footer_scripts", but that isn't a real variable, and produces the following error:  Catchable fatal error: Argument 2 passed to League\Plates\Template\Template::layout() must be of the type array, null given, called in /cdl/www/vhosts/dev.cdl.ucf.edu/udoit/templates/udoit.php on line 12 and defined in /cdl/www/vhosts/dev.cdl.ucf.edu/udoit/vendor/league/plates/src/Template/Template.php on line 149

I tried changing it to "$settings['footer_scripts']", which got me to the UDOIT interface without error, but then none of the javascript loaded.

PR #48, opened 2015-10-23
UDOIT now acts independent of hard coded refere

techrangerfj:


bagofarms:
The referer rest doesn't need to exist anymore.  The main point of that issue is to have the APIs work off of the value given by Canvas.  I'm closing this pull request.  Create a new one when the APIs use the value given by Canvas.

Issue #52, opened 2015-11-01
Udoit submit form doesn't handle error case


iturgeon:
In my work on adding heroku support I noticed the udoit submit doesn't handle error state when submitting to progress.php.

If progress.php errors, the interval just keeps running silently without updating the user.

PR #53, opened 2015-11-01
handle process.php and progress.php errors in ajax

iturgeon:
* Stops the progress checker when there's an error in either script.
* Simplifies the progress display update to reduce redundancy and require less maintenance.
* Globalizes the progress checker interval holder.
* Makes sure the interval is stopped when starting a new one.


iturgeon:
handles issue #52

PR #70, opened 2015-11-30
Fix for issue 69, base_url not being saved properly.

bagofarms:


bagofarms:
Also, i updated the composer lock file and composer.phar

Issue #76, opened 2015-12-04
Focus misplaced on expanding/collapsing view source


robfentress:
On the report page, clicking the link "View the source of this issue" causes the text of the link to be changed to "Close this view" and the source to be displayed, but focus then becomes indeterminate.  For instance, pressing tab after this moves focus to the "Scan Course" tab, which is far above where the user was in the page.  What should happen is that focus should be placed on "Close this view".  Probably even better would be for the text of the link to say "Hide source of this issue" so that it was clearer that this is a toggle.  

Similarly, pressing "Close this view" also causes focus to be indeterminate when it should stay on the link.

robfentress:
The second and third examples here seem to work for disclosure widgets:
http://codepen.io/afercia/pen/dogjjv

However, it would probably be better if this used <code>&lt;details&gt;</code> and <code>&lt;summary&gt;</code>.  I can't get that to work consistently yet, though.

Issue #74, opened 2015-12-01
Add throbber to "Safe report as PDF" button while the PDF is generating


bagofarms:

cooperfellows:
This does not work with the Heroku deployment, it results in a 404 when looking for the progress file:
![image](https://cloud.githubusercontent.com/assets/6798541/14787858/510e8dd4-0ad3-11e6-971c-cf1c6846cb07.png)

So the spinner never finishes, as can be seen in this video:
- http://screencast.com/t/XK1aZEdsfIML

Why is the `parsePdfProgress.php` file located in the lib directory and not in the public directory somewhere?

Issue #73, opened 2015-12-01
Remove empty numbered lists from errors and suggestions in generated PDF


bagofarms:
![screenshot 2015-12-01 09 05 43](https://cloud.githubusercontent.com/assets/1225334/11502779/cece4a84-980a-11e5-825b-23b3e91537d6.png)

jwbowles:
This is an easy fix. Just add result_html.find('.fix-success').remove(); in default.js

Issue #65, opened 2015-11-12
Change pNotUsedAsHeader to a Suggestion


bagofarms:
pNotUsedAsHeader is an error right now, but it should be a suggestion.  We are not sure of the content creator's intent with a bolded line of text, so it makes more sense as a suggestion.  Plus, it's already listed as a suggestion in "What does UDOIT look for?"

PR #82, opened 2015-12-22
Issues #65 #73 #74 #76 addressed

techrangerfj:


bagofarms:
Please pull master and resolve any issues.  This pull request is next on my list.

bagofarms:
We're including jscolor through Bower now, so please revert the change you made to udoit.php and remove the jscolor library from the repository.  Then, run `php composer.phar install`, which should run Bower after updating the composer libraries.

The PDF throbber does not stop after the PDF has downloaded.  It looks like it's because parsePDFprogress.php is not in the public/lib folder where the frontend is looking for it.

The empty numbered lists for #73 have been resolved, but now the PDFs don't have any CSS applied to them.  Here's what it used to look like:

screenshot 2016-02-05 08 54 34

and here's what it looks like now:

screenshot 2016-02-05 08 55 15

techrangerfj:
This one is good to go @bagofarms

Issue #91, opened 2016-02-10
Re-Implement styling to exported PDF report


techrangerfj:
The PDFs don't have any CSS applied to them.  Here's what it used to look like:

<img width="874" alt="screenshot 2016-02-05 08 54 34" src="https://cloud.githubusercontent.com/assets/1225334/12848158/1dd1500e-cbe6-11e5-8285-43b55c7af760.png">

and here's what it looks like now:

<img width="866" alt="screenshot 2016-02-05 08 55 15" src="https://cloud.githubusercontent.com/assets/1225334/12848178/32b9d964-cbe6-11e5-8ed7-9b017eccffc3.png">

techrangerfj:
Turns out when the files directory was changed, it caused the stylesheet for the pdf to be incorrectly be referenced in parsePdf.php

PR #92, opened 2016-02-22
Issue #91 addressed - PDF Styles restored

techrangerfj:
Turns out when the files directory was changed, it caused the stylesheet for the pdf to be incorrectly be referenced in parsePdf.php

Issue now addressed


techrangerfj:
@bagofarms good to go

Issue #117, opened 2016-03-24
Remove feedback prompt from templates/udoit.php


bagofarms:

emilysachs:
Resolved in this commit https://github.com/ucfopen/UDOIT/commit/4815495455e78542714e076b77e893408d590f9e#diff-0c2c7c4d879752ee30b98fc11328c9fd

Issue #116, opened 2016-03-24
Add disclaimer to default welcome text


bagofarms:
Add a disclaimer to the bottom of the default welcome text that reads:

Please Note:  This tool is meant to be used as a guide, not a certification.  It only checks for common accessibility issues, and is not comprehensive; a clean report in U**DO**IT does not necessarily mean that your course is fully accessible.  Likewise, the tool may indicate a possible accessibility issue where one does not exist.

emilysachs:
Resolved in this commit https://github.com/ucfopen/UDOIT/commit/4815495455e78542714e076b77e893408d590f9e#diff-0c2c7c4d879752ee30b98fc11328c9fd

PR #118, opened 2016-03-24
Issue #116 & #117 Implemented

techrangerfj:


bagofarms:
500 error on load.  Add a semicolon at the end of this line:  https://github.com/ucfcdl/UDOIT/blob/UDOIT_issue116_117/config/settings.php#L26

Issue #136, opened 2016-04-22
UFIXIT Color change not working with magic quotes


adrian-goetz:
The color change feature does not successfully update blocks of text with magic quotes.

emilysachs:
Found a similar issue with blocks of code enclosed by 'code' tags. The HTML entity conversion seems to be an underlying issue.

PR #146, opened 2016-05-04
Issue #136 - Ufixit now enabled on blocks of text including magic quotes and other special characters 

feneljoseph:
Issue states: The color change feature does not successfully update blocks of text with magic quotes.


techrangerfj:
@bagofarms I was wrong, it was this pull request that was being reviewed earlier

PR #152, opened 2016-05-05
Mesa CTL security updates

jwbowles:
Filter inputs to protect against XSS attacks. Clean up PHP warnings if Google Analytics tracking ID is left blank in the local config.


iturgeon:
:+1:  on the `filter_input_array` and `htmlspecialchars` usage

Some work needs to be done to merge this in and match the existing code/style.

Escaping for output is important - thanks for doing this.  Other portions of UDOIT are using templates that have this functionality built in.  I know they aren't using the templates now, but I'd suggest we fold these scripts that have display & logic mixed together into the template system and use the it's build in escape methods.  Cleaner, less redundancy.

I can attack this today - would it be stepping on your toes to run with the suggestions I have based on your work?

jwbowles:
That sound great. I know using a template system is significantly better,
but i didn't want to add yet another library to the project. I didn't
realize one was already in place! We have an application that scans the
repository to evaluate the code for vulnerabilities. So when you are done I
will perform another scan to see if we're still eliminating the XSS
vulnerabilities.

Thanks for all your work on this!

James

public/index.php
```diff
@@ -25,12 +25,16 @@
 ini_set("display_errors", 1);
 session_start();
 header('Content-Type: text/html; charset=utf-8');
+//Sanitize $_POST parameters
+$post_input  = filter_input_array(INPUT_POST, FILTER_SANITIZE_STRING);
 
 $templates = new League\Plates\Engine('../templates');
 
+$post_input['oauth_consumer_key'] = filter_input(INPUT_POST,'oauth_consumer_key',FILTER_SANITIZE_NUMBER_INT);
```
iturgeon: It looks like oauth_consumer_keys are not limited to integers - in fact I don't think the spec defines the character set at all

jwbowles: I can see about contacting Instructure to get a specific character set. If not we can just use the generic FILTER_SANITIZE_STRING to get at least the baseline sanitization.

iturgeon:
I got pretty deep into the version of this that uses the template engine.  The results page is pretty gnarly, got caught up in breaking it down into manageable parts.  That said I think combining the templates and these changes will work out.  I just need a little more time.

public/index.php
iturgeon: I'm falling back to FILTER_SANITIZE_STRING for now.  I know with LTI tools the consumer keys are pasted in by the user - so if that's the case here, they'll be just about anything.

iturgeon:
Ok, I buttoned up the pull request with all the things I was recommending.  I went a little nuts on it, but I think the project is a little cleaner now.  Anyway its over at #154 now.

bagofarms:
Closing in favor of #154

PR #169, opened 2016-06-08
Issue161 - Input Field Font Color within cssTextHasContrast now variable

techrangerfj:
Font color changed to either white (#fff) or black (#000) depending on background color luminosity


bagofarms:
These changes were merged in a different request.

Issue #149, opened 2016-05-05
Feature Request - Ability To Rename Tool In Canvas


cooperfellows:
One of the request I received from my eLearning team was to rename the tool when it's placed in the navigation. When I deployed the initial version of this, I simply changed the XML configuration file.

Now that we are moving to the Heroku deployment version, this is not going to be possible, since I want to keep the Git branch I deploy from consistent with the branches as they are developed.

I'd like to add the tool name as a configuration option. Is this something you would be OK supporting?

PR #150, opened 2016-05-05
Add configuration variable for app name in Canvas nav menu

cooperfellows:
Provide the ability to change the app name from "UDOIT" to any arbitrary text value. Relates to #149

PR #199, opened 2016-12-21
Add configuration variable for app name in Canvas

cooperfellows:
This is a re-roll of PR #150, which resolved issues #149. Let me know if anything looks out of place.

### DIFF
--- a/app.json
+++ b/app.json
@@ -39,6 +39,10 @@
     "USE_HEROKU_CONFIG": {
       "description": "needed to use the Heroku configuration",
       "value": "true"
+    },
+    "CANVAS_NAV_ITEM_NAME" : {
+      "description": "The text displayed in the canvas navigation menu to launch this app",
+      "value": "UDOIT"
     }
   },
   "addons": [
--- a/config/localConfig.template.php
+++ b/config/localConfig.template.php
@@ -12,6 +12,9 @@ $oauth2_uri = '';
 /* Disable headings check character count */
 $doc_length = '1500';
 
+/* Tool name for display in Canvas Navigation */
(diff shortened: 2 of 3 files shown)


bagofarms:
I can't get this to work on a non-Heroku install.  I even replaced the reference to the environment variable in localConfig.php to a hard-coded string with no luck.  It looks like udoit.xml.php isn't loading the settings.php or localConfig.php files, so $canvas_nav_item_name is always undefined.

cooperfellows:
I see what's missing from that file, it's something like this:

```php
// Get tool configuration values from the correct config file
if (getenv('USE_HEROKU_CONFIG'))
{
	try {
		require_once('../config/herokuConfig.php');	
	} catch(Exception $e) {
		//$logger = new \Monolog\Logger('udoit');
	}
	
	if(!empty($_SERVER['HTTP_X_FORWARDED_PROTO']) && $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https') {
		$_SERVER['HTTPS'] = 'on';
	}
}
else
{
	try {
		require_once('../config/localConfig.php');	
	} catch (Exception $e){
		// Log using Monolog
	}	
}
```

I'd like to log a failure if we either of these requires fails, but im not sure how to use Monolog, any examples?

bagofarms:
config/settings.php already handles determining whether you are in Heroku or a traditional install, so all you need to do is load it.  See https://github.com/ucfopen/UDOIT/blob/master/config/settings.php#L9

cooperfellows:
Thanks @bagofarms, I obviously haven't looked at the big re-write close enough, apologies for that.

Issue #100, opened 2016-02-24
pNotUsedAsHeader triggered even if role="heading" is used


robfentress:
The pNotUsedAsHeader rule is triggered, even if role="heading" has been used to indicate that a styled `<p>` tag is to serve as a heading.  For example, this throws a warning:

```
<p role="heading" aria-level="1"><strong><span style="font-size: 24pt;">I'm a level 1 heading</span></strong></p>
<div role="heading" aria-level="2"><strong><span style="font-size: 18pt;">I'm a level 2 heading</span></strong></div>
<p>I suspect this will trigger a warning about adding headings, but we'll see.</p>
<p><img src="/courses/45/files/5563/preview" alt="The W3C" width="72" height="48" data-api-endpoint="https://vt.test.instructure.com/api/v1/courses/45/files/5563" data-api-returntype="File" /></p>
<p role="heading" aria-level="2"><strong><span style="font-size: 18pt;">I'm another level 2 heading</span></strong></p>
<p>But I must explain to you how all this mistaken idea of denouncing pleasure and praising pain was born and I will give you a complete account of the system, and expound the actual teachings of the great explorer of the truth, the master-builder of human happiness. No one rejects, dislikes, or avoids pleasure itself, because it is pleasure, but because those who do not know how to pursue pleasure rationally encounter consequences that are extremely painful. Nor again is there anyone who loves or pursues or desires to obtain pain of itself, because it is pain, but because occasionally circumstances occur in which toil and pain can procure him some great pleasure. To take a trivial example, which of us ever undertakes laborious physical exercise, except to obtain some advantage from it? But who has any right to find fault with a man who chooses to enjoy a pleasure that has no annoying consequences, or one who avoids a pain that produces no resultant pleasure? On the other hand, we denounce with righteous indignation and dislike men who are so beguiled and demoralized by the charms of pleasure of the moment, so blinded by desire, that they cannot foresee.</p>
[...]

Issue #99, opened 2016-02-24
noHeadings triggered even if role="heading" is set


robfentress:
The noHeadings rule is triggered, even if role="heading" has been used to indicate that an element is to serve as a heading.  For example, this throws a warning:

```
<div role="heading" aria-level="1">I'm a level 1 heading</div>
<div role="heading" aria-level="2">I'm a level 2 heading</div>
<p>I suspect this will trigger a warning about adding headings, but we'll see.</p>
<p><img src="x.jpg" alt="The W3C" width="72" height="48" /></p>
<div role="heading" aria-level="2">I'm another level 2 heading</div>
<p>But I must explain to you how all this mistaken idea of denouncing pleasure and praising pain was born and I will give you a complete account of the system, and expound the actual teachings of the great explorer of the truth, the master-builder of human happiness. No one rejects, dislikes, or avoids pleasure itself, because it is pleasure, but because those who do not know how to pursue pleasure rationally encounter consequences that are extremely painful. Nor again is there anyone who loves or pursues or desires to obtain pain of itself, because it is pain, but because occasionally circumstances occur in which toil and pain can procure him some great pleasure. To take a trivial example, which of us ever undertakes laborious physical exercise, except to obtain some advantage from it? But who has any right to find fault with a man who chooses to enjoy a pleasure that has no annoying consequences, or one who avoids a pain that produces no resultant pleasure? On the other hand, we denounce with righteous indignation and dislike men who are so beguiled and demoralized by the charms of pleasure of the moment, so blinded by desire, that they cannot foresee.</p>
<p>But I must explain to you how all this mistaken idea of denouncing pleasure and praising pain was born and I will give you a complete account of the system, and expound the actual teachings of the great explorer of the truth, the master-builder of human happiness. No one rejects, dislikes, or avoids pleasure itself, because it is pleasure, but because those who do not know how to pursue pleasure rationally encounter consequences that are extremely painful. Nor again is there anyone who loves or pursues or desires to obtain pain of itself, because it is pain, but because occasionally circumstances occur in which toil and pain can procure him some great pleasure. To take a trivial example, which of us ever undertakes laborious physical exercise, except to obtain some advantage from it?</p>
<div role="heading" aria-level="2">I'm also a level 2 heading</div>
[...]

PR #220, opened 2017-01-23
Add support for p and div tags with role="heading"

cooperfellows:
Solves #99 
Solves #100 

This will probably need a good review from others. Right now we are only checking for div and p tags that could have a roles of header. I imagine there are other block level elements that we may want to add in the future. Hence, the array of elements on line 2928, adding more elements to this search should be nice and easy.

### DIFF
--- a/lib/quail/quail/common/accessibility_tests.php
+++ b/lib/quail/quail/common/accessibility_tests.php
@@ -2908,8 +2908,10 @@ class noHeadings extends quailTest
 
 		if (strlen($document_string) > $doc_length){
 
-			$no_headings = 0;
+			// Assume we have headings, look for all possible heading types
+			$no_headings = false;
 
+			// Traditional heading elements
 			if (!$this->getAllElements('h1')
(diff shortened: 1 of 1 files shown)


lib/quail/quail/common/accessibility_tests.php
```diff
@@ -2908,17 +2908,59 @@ function check()
 
 		if (strlen($document_string) > $doc_length){
 
-			$no_headings = 0;
+			// Assume we have headings, look for all possible heading types
```
cooperfellows: This comment should probably read "Assume we **don't** have headings, look for all possible heading types".
Think that was just a typo on my part

cooperfellows:
@emilysachs Hold off on reviewing this one yet, I think I found a bug but want to review closer tomorrow. Sorry for the backpedaling.

bagofarms:
There are still issues with this PR, so I'm going to close it.  Feel free to reopen it if the issues are resolved.

Issue #191, opened 2016-09-09
Fix alt text doesn't work


jwbowles:
I noticed that the fixAltText function in Ufixit.php is returning the img tag without the new alt text added properly. After investigating further it appears that
`$remove_attr = preg_replace("/ data-api-endpoint.+?>/", "", $fixed_img);
        $fixed_img = $remove_attr;
`
strips the data-api-endpoint tag as well as any text following that tag. This happens to include the alt tag because it is appended to the end of the new HTML string. This just recently started occurring without any modification to the code on our part.

jwbowles:
Here is a fix I developed to address this issue:
`public function fixAltText($error_html, $new_content, $submitting_again = false)
    {
        $this->dom->loadHTML('<?xml encoding="utf-8" ?>' . $error_html);

        $imgs = $this->dom->getElementsByTagName('img');
        $fixed_img = null;

        foreach ($imgs as $img) {
            $img->setAttribute('alt', $new_content);
            $img->removeAttribute('data-api-endpoint');
            $img->removeAttribute('data-api-returntype'); 
            $fixed_img = $this->dom->saveHTML($img);
        }

        $fixed_img = trim($fixed_img,">");

        return $fixed_img;
    }`

This will remove data-api-endpoint and data-api-returntype attributes and then trim the trailing ">" from the tag so that the additional attributes Canvas requires will be appended to the end of the image tag.

bagofarms:
Sorry i didn't get to review this issues sooner.  Thank you so much for providing the code.  I know we're backed-up on pull requests right now, but if you could submit one, that would be great!

cooperfellows:
I've been looking a lot at this issue the past two days, hoping to have a discussion about the best way forward here. @jwbowles is right, the current system doesn't work, as the alt tag is removed by the regex as pointed out.

I started with his proposed fix from PR #192, but found that none of the text was replaced. Through some other trial and error ([you can see that history here](https://github.com/ucfopen/UDOIT/compare/master...cooperfellows:UDOIT_issue191)) I managed to get it working correctly. I cleaned up all the debug commits and the only change that we really need to make is [seen here](https://github.com/ucfopen/UDOIT/compare/master...cooperfellows:UDOIT_issue191_clean).

However, I am wondering why we are trying to remove those attributes in the first place? I understand not wanting to introduce extra code back into a page (or syllabus in this case), but this is not the way to do it.

[...]

PR #192, opened 2016-09-16
Issue191 - Fixing "fixAltText"

jwbowles:
Update fixAltText to specifically target the data-api attributes that are added by Canvas to the image  tag without also removing the alt attribute.

PR #223, opened 2017-01-24
Ensure fixAltText Method Works

cooperfellows:
Related to #191 
Replaces #192 

[Further discussion](https://github.com/ucfopen/UDOIT/issues/191#issuecomment-274875490) is needed on issue #191


emilysachs:
Fixes #191

Issue #203, opened 2016-12-22
UFIXIT - Color Contrast:  When background color is set, background and foreground colors get placed in the opposite fields.


bagofarms:
![screenshot 2016-12-22 11 03 21](https://cloud.githubusercontent.com/assets/1225334/21431705/5bd40e62-c836-11e6-8be3-fd91dda359e6.png)

PR #224, opened 2017-01-25
Fix color contrast issue

cooperfellows:
The background color was being passed to the server and interpreted as
the forground color by mistake. The index of the "newcontent" field name
needed to be switched.

This solves #203

### DIFF
--- a/templates/partials/result_item/contrast.php
+++ b/templates/partials/result_item/contrast.php
@@ -26,12 +26,12 @@
 	<div class="left">
 		<div class="form-group no-margin margin-bottom">
 			<?php if ( isset($group_item->back_color) ): ?>
-				<label for="newcontent[1]">Replace Background Color <?= $group_item->back_color; ?></label>
-				<input class="color {hash:true,caps:false} form-control back-color" type="text" name="newcontent[1]" value="<?= $group_item->back_color; ?>" placeholder="Replacement for Background Color <?= $group_item->back_color; ?>">
+				<label for="newcontent[0]">Replace Background Color <?= $group_item->back_color; ?></label>
+				<input class="color {hash:true,caps:false} form-control back-color" type="text" name="newcontent[0]" value="<?= $group_item->back_color; ?>" placeholder="Replacement for Background Color <?= $group_item->back_color; ?>">
 			<?php endif; ?>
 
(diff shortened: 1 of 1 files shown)


cooperfellows:
@emilysachs Are you sure we want to change the library code? The only reason that is "backwards" in reporter.static.php is because of the way we are sending the data.

It would be my preference, but I will obviously defer to your team, to change our code to work with the Library as it exists. Does that make sense?

emilysachs:
Let me take another look! I only tried altering the library because when I pulled in your branch I was still seeing an issue. Perhaps there's something else affecting it.

cooperfellows:
If you continue to see issues with the branch (without changing the reporter code) I will take another look. 

I've been deploying each PR branch to Heroku and doing manual testing each time, so I could have missed something.

emilysachs:
@cooperfellows It looks like the problem actually has to do with the order of the colors in the attributes. When I test it on a span with "background-color: #333333; color: #003366;" it works, but "color: #003366; background-color: #333333;" shows up as flipped in UDOIT. So we'll need to add in a way to put the attributes in the right order that matches the way our code is referencing them. That being said, while testing I also noticed that the contrast test doesn't recognize a case like 
`<span style="background-color: #333333;"><span style="color: #003366;">test</span></span>`
correctly and treats each span tag as a separate problem if it triggers a contrast error by itself. Not sure yet if it's worth opening a new issue for this, or if it might be an easy fix.

cooperfellows:
Good eyes @emilysachs, The second issue that you raised, nested contrast issues, are going to be harder to solve I imagine, that may warrant a new issue.

The first one (attribute ordering causing a problem) makes total sense, and I think should be resolved as a part of this issue. I'm not sure how much time I can give that this week (or next honestly), but I will keep progress posted to this PR.

emilysachs:
@cooperfellows That's what I was guessing, I'll go ahead and open a new issue for it. I'm just about to leave for today but I'll take a shot at the attribute ordering tomorrow if that's alright.

cooperfellows:
Go ahead and do whatever you need to @emilysachs, I won't be able to look at this until mid next week at the earliest.

emilysachs:
Attribute ordering resolved here https://github.com/ucfopen/UDOIT/pull/237

bagofarms:
Do we need this PR anymore, or does PR #237 fix it?

Also, just a quick note:  We have already heavily modified the Quail code, and the PHP version is not maintained anymore, so I'm comfortable with modifying it directly for things like this in the future.

emilysachs:
@bagofarms I pulled the changes from this PR into 237 before adding the extra fixes on top of it, so I believe we can close this one.

Issue #173, opened 2016-06-29
Color contrast insufficient in several areas


robfentress:
Color contrast for `.badge-error` and `.btn-success` is insufficient, as is the white on grey text of the labels for panels.  I'm not sure where that is in the code, but it is in the upper left hand corner of the panels.

Here is the preview presented by the [Color Contrast Analyzer Chrome extension](https://chrome.google.com/webstore/detail/color-contrast-analyzer/dagdlcijhfbmgkjokkjicnnfimlebcll?hl=en), with 3:1 contrast ratio being tested for.  Anywhere where there isn't an outline around the text, there is insufficient contrast:

![contrast-wcag-aa-large](https://cloud.githubusercontent.com/assets/9356499/16451647/4bef6a30-3dd2-11e6-84a2-884c3c1cbfe8.png)

PR #225, opened 2017-01-25
Fix contrast ratio issues with UDOIT Elements

cooperfellows:
Several of the UDOIT elements do not meet the required contrast ratio
for WCAG2.0 (AA). The changes in this commit resolve all color contrast
issues as reported by the Chrome Developer Tool Accessibility audit
system
(https://chrome.google.com/webstore/detail/accessibility-developer-t/fpkknkljclfencbdbgkenhalefipecmb?hl=en)

It also fixes what I imagine was a bug, where the "UDOIT" green buttons
were turning green on hover due to line 62 being too generic.

Solves #173

### DIFF
--- a/public/assets/css/main.css
+++ b/public/assets/css/main.css
@@ -59,7 +59,8 @@ body { padding: 20px 0 0; }
 	margin: 0;
 }
 
-.list-group-item .btn:hover{
+.list-group-item .viewError.btn:hover,
+.list-group-item .closeError.btn:hover{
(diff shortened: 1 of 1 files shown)


cooperfellows:
Happy to make these changes, I'm not much in the way of a designer, thanks for the input.

emilysachs:
Requested changes added here https://github.com/ucfopen/UDOIT/pull/276

PR #226, opened 2017-01-26
Udoit issue168 - Add Unique ID to Form Elements

### LINKED ISSUE #168: U FIX IT combobox for table headers lacks label
When you attempt to use U FIX IT to add headers to a table (`tableDataShouldHaveTh`), you are presented with a combobox that is not labeled. It should either be associated with a `<label>` tag (preferable) or use `aria-label` or `aria-labelledby` to give the `<select>` box an accessible name.
screen shot 2016-06-08 at 3 27 38 pm

cooperfellows:
This started off as me trying to resolve #168. Looking at the new templating system and the form elements that are output, it became clear that the missing label identified in #168 was indicative of a more widespread problem. Most of the form input fields we use during the UFIXIT phase are completely missing labels, or have a label element, but that label element does not actually link to the form input because those form inputs don't have unique ID's.

To resolve this, in https://github.com/cooperfellows/UDOIT/commit/5227a48a66614cf1d17bdc7105bbc40c71ebb2d7, I am now passing the unique id that is generated and placed on the `<li>` item (this is used to enable the expand/collapse of each item in the report).

Since we have a somewhat unique ID already, we are able to build on top of it with a simple modifier:
```php
id="<?= $this->e($item_id); ?>-input"
```
or for the color contrast system
```php
 id="<?= $this->e($item_id);?>-bgcolor"
```

Then a label is added that hooks up to those items:
```php
<label for="<?= $this->e($item_id); ?>-bgcolor">
```
or
```php
<label for="<?= $this->e($item_id); ?>-input"
```

I _believe_ this handles all the form inputs when reviewing proposed fixes. I identified these by using the [Chrome Developer Tools A11Y Audit](https://chrome.google.com/webstore/detail/accessibility-developer-t/fpkknkljclfencbdbgkenhalefipecmb?hl=en).

There is still a good chunk of work to do to push this whole tool toward WCAG2.0 compliance. For instance, the expand and collapse functionality we have, not only for the sections (Page 1, Page 2, etc...), but also for the "View Issue Source", should be wired up with aria attributes to comply with the [spec for accordions](https://www.w3.org/TR/wai-aria-practices-1.1/#accordion). That seemed like a lot of work to put into this PR, so I plan to create a new issue for that and submit fixes for that later.

I see two options moving forward on that front, build the aria labeling system into our existing templates, or switch our templates around to use an existing aria compliant accordion. [I've used one on some other projects](https://github.com/cooperfellows/a11y-accordion) with success. It is a relatively small library (8.33KB uncompressed, or 3.41KB minified), but it is another library. So I'm open to suggestions about the best way forward there.

### DIFF
--- a/templates/partials/result_item/contrast.php
+++ b/templates/partials/result_item/contrast.php
@@ -26,12 +26,12 @@
 	<div class="left">
 		<div class="form-group no-margin margin-bottom">
 			<?php if ( isset($group_item->back_color) ): ?>
-				<label for="newcontent[1]">Replace Background Color <?= $group_item->back_color; ?></label>
-				<input class="color {hash:true,caps:false} form-control back-color" type="text" name="newcontent[1]" value="<?= $group_item->back_color; ?>" placeholder="Replacement for Background Color <?= $group_item->back_color; ?>">
+				<label for="<?= $this->e($item_id); ?>-bgcolor">Replace Background Color <?= $group_item->back_color; ?></label>
+				<input class="color {hash:true,caps:false} form-control back-color" type="text" name="newcontent[1]" value="<?= $group_item->back_color; ?>" placeholder="Replacement for Background Color <?= $group_item->back_color; ?>" id="<?= $this->e($item_id);-bgcolor?>">
 			<?php endif; ?>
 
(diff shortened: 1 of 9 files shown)


iturgeon:
Yea, generally good direction I think.  Theres a lot of jquery dom searching I'm not excited about that is related to this.  Using the item_id's sounds useful, without looking myself are those unique enough?

cooperfellows:
@iturgeon The item_id as generated now is unique at the `<li>` level, but if we want to use that to label other items (panel body in the accordion, inputs, etc...) it obviously needs to be appended with new data to remain unique, you can see in b046214 that I had screwed that up the first time around.

Issue #235, opened 2017-02-08
Color contrast test doesn't recognize nested styling


emilysachs:
Cases like `<span style="background-color: #333333;"><span style="color: #003366;">test</span></span>` are handled as two separate contrast checks.

PR #243, opened 2017-03-01
Color contrast and text emphasis UFIXIT fixes

emilysachs:
This started out as a solution for issue #235 and #203. The process of fixing and testing those changes subsequently uncovered a lot of things that weren't working in both the UFIXIT code and the quail library.

- Added warning to cssTextHasContrast description about nested styling
- Updated fixCssColor() to strip original colors and text emphasis and then add in fixed colors and text emphasis with respect to inheritance
- Added font-weight and font-style logic to fixCssEmphasize() to support inheritance 
- Wrote new XPath query for cssTextHasContrast and cssTextStyleEmphasize to select only the relevant text nodes and ignore repeats of identical parent nodes, preventing excessive UFIXITs that overwrite each other
- Adjusted logic of cssTextHasContrast to account for inheritance (by fixing broken walkUpTreeForInheritance and getNodeStyle functions in quail) and return the proper background color, text color, font style, and font weight to the UFIXIT view
- Fixed cssTextEmphasize to include italics
- Added logic to detect strong and em elements as bold and italic
- Fixed error in quail color parsing that returned no value for 3 character hex codes, added else case to default to black text when not 3 or 6 characters
- Added color converting step to cssTextHasContrast so name value colors appear in UFIXIT preview
- Simplified logic in reporter.static.php to account for new return format from cssTextHasContrast
- Adjusted javascript to detect changes in bold and italics checkboxes and fixed problem with displaying the changed color in real time
- Adjusted UFIXIT templates to pull current values for bold and italics checkboxes on load and display proper styling in preview box


emilysachs:
Working on updating the test cases to accommodate new parameters for UFIXIT functions

Issue #216, opened 2017-01-19
Add message to "Unscannable" area suggesting the use of Pages instead


bagofarms:
The message should be configurable (and you can turn it off), but the default should say:
Consider converting these documents to Pages, since they are easier to update and generally more accessible.

bagofarms:
Consider adding a reference to mobile devices.

PR #248, opened 2017-03-08
Adding unscannable suggestion

emilysachs:
Issue #216 and fixing unclosed bold tags


emilysachs:
Should address issue #258

Issue #252, opened 2017-04-03
Only display current alternative text for source in report


rebelaide:
Instead of showing the entire source of the image tag, only display the alt source. This would be less confusing for non-technical users.

Current:

Line 4: <img src="https://webcourses.ucf.edu/courses/1003256/files/35767586/preview?verifier=rIIt0qB17kBolEwXOHpysv2iBuUKYiHv643CVihi" **alt="kids-tech-revised.jpg"** data-api-endpoint="https://webcourses.ucf.edu/api/v1/courses/1003256/files/35767586" data-api-returntype="File">

Proposed:

 alt="kids-tech-revised.jpg"

mkattut:
I like this idea, but what about images that have null alt text?

rebelaide:
Null is displayed as an empty alt attribute in the source.

Current:

Line 1: <img src="https://webcourses.ucf.edu/courses/1003256/files/35767586/download?verifier=rIIt0qB17kBolEwXOHpysv2iBuUKYiHv643CVihi" **alt=""** data-api-endpoint="https://webcourses.ucf.edu/api/v1/courses/1003256/files/35767586" data-api-returntype="File"> 

I think just displaying it as empty would be sufficient. 

Proposed:

alt = ""

The image would be still displayed in the preview box above the source code.

emilysachs:
(Discussed in person)
Going to test out adding a similarly-styled box with just the alt text next to the Preview and Source boxes.

bagofarms:
This feature will be present in 3.0.0 upon its release.

PR #254, opened 2017-04-04
Adding Alt Text under view source for appropriate tests

emilysachs:
Makes visibility of alt text better for non-technical users - Fixes issue #252


bagofarms:
This functionality will be present in 3.0.0

Issue #260, opened 2017-04-18
Remove list of issues from "Consider adding headings" issue.


bagofarms:
From Michelle Tuten at Clemson:

> One more thing I noticed is that the Consider adding headings to your document to create more structure error has a collapsible area where the view source and U Fix IT button would normally be that instead only contains a list item number. Is it possible to forego the collapsible area if it contains no content?

![wall of text](https://cloud.githubusercontent.com/assets/1225334/25154107/38a7d292-245d-11e7-8b4b-c592595392b2.png)

Solution:
It should look like the "Collapsed" view in the image above.

PR #261, opened 2017-04-19
Don't show list item for noHeadings error

emilysachs:
Fixes issue #260 
Enclosed code that renders list of errors in "if ($type != "noHeadings")" to not show a list for this specific error


bagofarms:
I'll take a look at how this can be merged into the current codebase.  This is still an issue.  I'll take a look at what the "content length" suggestion is doing to avoid this.

jasonmatos23-git:
Manually merged the 'if ($type != "noHeadings")' conditional from 'results_items.php' into the dev branch's version. Tested and it fixes the issue!

AlanFCMV (approved):
Tested myself and issue is fixed

Issue #102, opened 2016-02-24
objectMustContainText rule triggered even if aria-labelledby used


robfentress:
The objectMustContainText rule triggered even if `aria-labelledby` is used to provide alternative text for the `<object>`.

For instance, this fails, when it shouldn't:

```
<p><object data="x.jpg" aria-labelledby="monkey"></object></p>
<div id="monkey">My monkey</div>
```

![screen shot 2016-02-24 at 4 13 48 pm](https://cloud.githubusercontent.com/assets/9356499/13301198/bf9ff1ee-db11-11e5-9873-325d71976571.png)

robfentress:
I've since found that, though, technically, using aria-labelledby for alternative text for objects is okay, the support for this is inconsistent, especially in Windows-Eyes and NVDA with Firefox, so the case for changing this one is diminished.

bagofarms:
@robfentress Would you mind if i put all of the aria-related issues as low priority?  My reasoning is that most HTML content in Canvas is created using the Rich Content Editor, and I doubt most content creators will be using aria attributes, since that would require modifying the HTML directly.

What do you think?

robfentress:
Correction again: When I used a real URL for the data source and set the width and height, NVDA worked fine, so I guess the issue comes down to whether you think Window-Eyes has enough market share to warrant inclusion.  For me, Window-Eyes crashes constantly, so I find it hard to believe it is used as much as the [WebAim survey](http://webaim.org/projects/screenreadersurvey6/) suggests, but I guess it is the best we've got.

robfentress:
I'd agree that most content added through the Rich Content Editor is unlikely to include aria attributes (though I've used them), but content uploaded through the Files tool might be more likely to.  I just hate the idea of giving false errors and wouldn't think it would be that hard to add a check for those attributes. . . but I'm not the one who has to do it.  And I guess it depends on what other fixes/improvements you've got in the pipeline as to what priority it should be given.

robfentress:
As to priority, reviewing the list of bugs I've reported, I think the following should be addressed, in this order, before any of the aria issues were addressed:

1. [Large text fails cssHasTextContrast incorrectly](https://github.com/ucfcdl/UDOIT/issues/86)
2. [UFIXIT offers to fix page in Files, but throws unhelpful error](https://github.com/ucfcdl/UDOIT/issues/108)
3. [Add input validation to UFIXIT](https://github.com/ucfcdl/UDOIT/issues/95)
[...]

Issue #101, opened 2016-02-24
objectMustContainText triggered even if aria-label used


robfentress:
The objectMustContainText rule triggered even if `aria-label` used to provide alternative text for the `<object>`.

![screen shot 2016-02-24 at 4 10 15 pm](https://cloud.githubusercontent.com/assets/9356499/13301125/5a4fc86e-db11-11e5-829f-148454cccdd7.png)

robfentress:
I've since found that, though, technically, using aria-label for alternative text for objects is okay, the support for this is inconsistent in Windows-Eyes and NVDA with Firefox, so the case for changing this one is diminished.

robfentress:
Correction again: When I used a real URL for the data source and set the width and height, NVDA worked fine, so I guess the issue comes down to whether you think Window-Eyes has enough market share to warrant inclusion.  For me, Window-Eyes crashes constantly, so I find it hard to believe it is used as much as the [WebAim survey](http://webaim.org/projects/screenreadersurvey6/) suggests, but I guess it is the best we've got.

jasonmatos23-git:
Closing because changes to 2.8 had been merged.

PR #267, opened 2017-05-09
Aria labels for objectMustContainText test

emilysachs:
Fixes issue #101 
Fixes issue #102


AlanFCMV (approved):
I tested empty objects, objects with empty aria labels, as well as objects with non-empty aria labels (of both types) and they all went as expected

Issue #279, opened 2017-05-19
Add Heroku background worker functionality


iturgeon:
Heroku times out on web requests (as it should), but that really breaks UDOIT during long api tasks.  

The free Heroku service appears to allow a background worker so free trial usage shouldn't be an obstruction moving forward.

Ideally, the requests from the front end are passed to the web api, then the web server would pass the api request tasks along to the background worker for long running tasks.  The existing polling functionality for the front end can continue to work the way it does - though it would need to decouple how that first long-running request works.

Background tasks need:
* [ ] set reasonable time out issues
* [ ] report error states appropriately to the application
* [x] fall back to non-heroku setups easily to aid existing local installs

iturgeon:
I'm interested in working on this issue. If there's anyone else that would like to provide feedback, feel free to speak up.

iturgeon:
Update:

I've been able to get a concept working on heroku with php and a background worker.

Now, I'm working on changes needed to UDOIT to support this functionality.  My current plan is to change the scan request from doing all the work

## Currently how it works

1. User selects (checkboxes) the areas to scan in course
2. Process.php runs a single long-running job/request that does all the scans sequentially.  This process is of unknown duration and updates a session variable to keep track of what section it's currently working on.  (This is *highly* susceptible to apache/nginx time-outs).
3. Meanwhile, the client keeps polling the server (2nd request) for the current state of the scan. Note heroku free servers don't handle more then 1 request at a time, so often these go unanswered until the process.php request returns.
4. Once the long-running process.php request finishes, it returns the results of the scan and the polling stops.

## The Plan

1. User selects (checkboxes) the areas to scan in course
2. Process.php splits each checkbox into a job to run and deposits them in the database, returning after the job requests are recorded. This request returns the id of the scan group.
3. Then, the client starts polling the server to check on the status of the scan group.
4. When the scans are completed, the polling stops and the client sends a new request for the results.

cooperfellows:
This sounds awesome Ian, we run on Heroku right now, and I've got our Sandbox Canvas account hooked up to separate install of UDOIT, so I'm happy to help test new releases as I have time. I will be up front, I amt totally swamped right now, so I don't be able to provide too much feedback, but enough to point out obvious problems.

Thanks for your work here, it sounds like a great addition.

iturgeon:
@cooperfellows the background worker is nearly complete, in the final stretches now.  Perhaps you'd be interested in testing it out?

[...]

PR #285, opened 2017-06-07
Removes Bower and NPM Requirement

### LINKED ISSUE #284: Remove Bower & NPM requirements
A combination of custom options should allow Composer to install the js git repo and place it in the required directory.

Doing so will dramatically reduce install complexity by removing the need for NPM and Bower.

iturgeon:
* No longer requires Bower & NPM to install jscolor.
* Adds Composer requirement of jscolor and a plugin that will move it to the correct directory post install.
* Updates composer.json & composer.lock to lock down versions of packages more accurately.
* Removes All bower config and references
* Removes nodejs buildpack from Heroku config
* Updates readme to remove Bower/Node docs

Fixes #284


iturgeon:
This is going to be a pre-req for #279

iturgeon:
Heroku installs faster now :)

Compressed before to 80.3M, now is 62.6M

iturgeon:
Verified jscolor still seems to work on heroku and everything installs.

PR #299, opened 2017-06-09
Changes report column name and adds migration script

iturgeon:
## UPGRADE NOTE

When upgrading from a previous version of UDOIT, `php composer.phar migrate` will need to be run.  If it's not run - the result shouldn't be bad.  They'll have an extra column in the reports table, and the reports that aren't migrated from disk to the table just can't be viewed.

## Changes

* Scripts now added to composer.json
* Script for moving disk based reports into the db added
* Column name for reports.file changed to reports.report_json
* Column creation updated to use MEDIUMTEXT on mysql
* scripts moved in to bin directory
* contributors moved to bottom of composer.json
* report reading no longer attemps to read from disk

fixes: #296

### DIFF
--- a/HEROKU.md
+++ b/HEROKU.md
@@ -119,7 +119,7 @@ If you need to check that the tables exist, you can connect to Postgres using so
 * `Select * from users;` or `Select * from reports;` will show you their contents
 * `\q` quits the psql terminal
 
-If needed, you can manually run the table creation script: `Heroku run php lib/db_create_tables.php`
+If needed, you can manually run the table creation script: `heroku run composer dbsetup`
 
 ## Table Schema
-The table schema can be found in [lib/db_create_tables.php](lib/db_create_tables.php)
+The table schema can be found in [bin/db_create_tables.php](bin/db_create_tables.php)
--- a/README.md
+++ b/README.md
@@ -96,10 +96,10 @@ Edit `config/localConfig.php`:
 To create the required tables, run the creation script below.  You'll need to complete the db steps above first.
 
 ```
-$ php lib/db_create_tables.php
+$ php composer.phar dbsetup
 ```
 
-The table schema can be found in [lib/db_create_tables.php](lib/db_create_tables.php)
+The table schema can be found in [bin/db_create_tables.php](bin/db_create_tables.php)
 
 ## Configuration and Setup
 If you didn't already make `config/localConfig.php` when you set up the database, do it now.
--- a/app.json
+++ b/app.json
@@ -54,6 +54,6 @@
     }
   ],
   "scripts": {
-    "postdeploy": "php lib/db_create_tables.php"
+    "postdeploy": "composer dbsetup"
   }
 }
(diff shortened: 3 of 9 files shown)


public/cached.php
```diff
@@ -22,7 +22,7 @@
 
 $dbh = include('../lib/db.php');
 // saves the report to the database
-$sth = $dbh->prepare(" SELECT * FROM {$db_reports_table} WHERE course_id = :courseid ORDER BY date_run DESC");
+$sth = $dbh->prepare(" SELECT id, date_run, errors, suggestions FROM {$db_reports_table} WHERE course_id = :courseid ORDER BY date_run DESC");
```
iturgeon: The columns were specified because this table now contains a huge report that's not needed for the view.

public/process.php
```diff
@@ -98,7 +93,6 @@
             Utils::exitWithPartialError('Error inserting report into database');
         }
 
-        $udoit_report = json_decode($udoit->getReport());
```
iturgeon: extraneous getReport() call

HEROKU.md
```diff
@@ -119,7 +119,7 @@ If you need to check that the tables exist, you can connect to Postgres using so
 * `Select * from users;` or `Select * from reports;` will show you their contents
 * `\q` quits the psql terminal
 
-If needed, you can manually run the table creation script: `Heroku run php lib/db_create_tables.php`
+If needed, you can manually run the table creation script: `heroku run composer dbsetup`
```
iturgeon: Switching to using composer scripts. so nice.

Composer is available on Heroku, no need to use `php composer.phar`

HEROKU.md
```diff
@@ -119,7 +119,7 @@ If you need to check that the tables exist, you can connect to Postgres using so
...
 
 ## Table Schema
-The table schema can be found in [lib/db_create_tables.php](lib/db_create_tables.php)
+The table schema can be found in [bin/db_create_tables.php](bin/db_create_tables.php)
```
iturgeon: Moved runnable command scripts into a bin folder

README.md
```diff
@@ -96,10 +96,10 @@ Edit `config/localConfig.php`:
 To create the required tables, run the creation script below.  You'll need to complete the db steps above first.
 
 ```
-$ php lib/db_create_tables.php
+$ php composer.phar dbsetup
```
iturgeon: I kept this using `composer.phar` for ease of use.

app.json
```diff
@@ -54,6 +54,6 @@
     }
   ],
   "scripts": {
-    "postdeploy": "php lib/db_create_tables.php"
+    "postdeploy": "composer dbsetup"
```
iturgeon: Again, this is in Heroku where composer is globally installed.

public/process.php
bagofarms: Turns out this isn't extraneous.  `parseReport.php` relies on this object existing.  I think this whole workflow could use a rethink.  I put this line back in dev/v2.2.0.

iturgeon: Ah, yes.

FWIW, #292 has some of this rethinking built in

Issue #317, opened 2017-12-20
Make file size warning value configurable


bagofarms:
The default value should be set to 50mb (originally created in #217 ).  This should be configurable as both a value in localConfig.php and as a Heroku environment variable.

PR #321, opened 2018-01-14
Issue 317 make max file size configurable

alexanderholman:
made max upload size configurable as per issue #317


alexanderholman:
sorry missed PR #320

Issue #327, opened 2018-02-09
Empty canvas_url values after migration


bagofarms:
After the migration that adds the canvas_url field to `users`, all of the existing values are empty.  This causes an error when UDOIT runs.  We could add a prompt to the migration process to request a URL to fill in the empty values.  We could add a variable to the localConfig called `$default_canvas_url` or something.

PR #330, opened 2018-02-15
Issue 327 Empty canvas_url fields now update automatically.

bagofarms:
We'll have to figure out how to deal with the same user id from multiple instances of Canvas in the future


bagofarms:
Fixes #327

Issue #353, opened 2018-03-27
Update PHP/Docker testing


seanlh:
PHP 54 and 55 are no longer supported. Running php composer.phar docker-test is returning some runtime errors, probably because some installation scripts need to be updated.

PR #354, opened 2018-03-27
Removed php54 and php55 support; fixed docker installation script for…

seanlh:
… older versions of PHP to use older versions of xdebug; fixed wrongly mounted volumes in docker-compose


seanlh:
Fixues issue #353

PR #362, opened 2018-04-11
Add quail support for parent and child strong tags with color emphasis

### LINKED ISSUE #339: Strong tag not detected with color emphasis
When a user creates colored, bolded text in the Canvas RCE, it comes out in this format:  `<strong><span style="color: #800000;">This is some text bolded with the RCE</span></strong>` or this format `<span style="color: #800000;"><strong>This is some text bolded with the RCE</strong></span>` depending on the order that they applied the styles.  UDOIT does not check for this correctly, instead relying on the `font-weight` in the style attribute of the span to determine if primary emphasis is applied in addition to the color (https://github.com/ucfopen/UDOIT/blob/master/lib/quail/quail/common/accessibility_tests.php#L1363)

seanlh:
Fixes #339

### DIFF
--- a/lib/quail/quail/common/accessibility_tests.php
+++ b/lib/quail/quail/common/accessibility_tests.php
@@ -1371,7 +1371,14 @@ class cssTextStyleEmphasize extends quailColorTest
 							}
 						}
 					}
-				} else if ($element->tagName === "strong") {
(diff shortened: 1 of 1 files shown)


Thomas-Whitaker:

PR #363, opened 2018-04-11
Make UFIXIT link text less restrictive

### LINKED ISSUE #340: Remove UFIXIT restriction that disallows link text found within URL
If you have a link to http://www.wikipedia.com/, and you try to set the link text to "Wikipedia" in UFIXIT, you will get the message "Please check that you've entered new text for this link and that it is descriptive".  There are many cases where the appropriate link text is found within the URL, so we should be less restrictive.  We should still detect if the link text contains URL elements or the entire URL itself.

Relevant line:  https://github.com/ucfopen/UDOIT/blob/v2.3.1/public/assets/js/default.js#L485

seanlh:
Fixes #340 

Instead of checking if link text is in URL, simply check if the link text contains
- http://
- www.
- .com


Thomas-Whitaker:

PR #364, opened 2018-04-11
Add error source DOM removal and reattachment functionality

### LINKED ISSUE #307: Video Playback continues when "View the source of the issue" is closed
(Hi all. New to Github. Please forgive and correct me, if I'm speaking where I'm not supposed to.)
When a course generates a "Synchronized captions should be provided for prerecorded web-based video" result, if someone clicks the "View the source of the issue" button and begins watching the video then closes the issue source, the video continues to play. 

I don't expect that this behavior is an issue for many, but I could see it as being an issue for someone with cognitive disabilities. Is there a way where the video playback can be paused or stopped when the "Close the issue" is pressed?

[Video Playback.zip](https://github.com/ucfopen/UDOIT/files/1341370/Video.Playback.zip)

seanlh:
Fixes #307


Thomas-Whitaker:

bagofarms:
If you view the source of the issue on a video, then view the source of the issue on a non-video, the video will be shown instead.

PR #385, opened 2018-10-02
Multitenant initial commit. 

webchuckweb:
Adds support for multitenancy (single UDOIT with multiple Canvas instances).

Adds a database table for institutes, management screen for adding/removing institutes, and generation of LTI consumer keys and secrets.


iturgeon:
I know @bagofarms will take a look as soon as he can.  Initially it looks like the CI test for code style is failing. 

If you run `composer run sniff-errors-only` it will pinpoint each of the errors from the code sniffer.

migrations/000_db_create_tables.php
```diff
@@ -26,6 +29,17 @@
             );
         ',
     ];
+
+    if ($use_multitenant) {
```
bagofarms: Database changes should be added to a new migration file, with the three digit number at the beginning of the filename incrementing by 1 each time.  In this case, it would be something like `004_add_multitenant`.

multitenant/form_handler.php
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Verify that .htaccess and .htpasswd files exist. Very basic security check, but it offers a reminder to the end user.
+ *
+ * @return bool
+ */
+function multitenant_check_security()
+{
+    $secure = false;
+
+    if (file_exists('.htaccess')) {
```
bagofarms: Using .htaccess and .htpasswd will definitely work for now, but I don't feel like it's a permanent solution.  Two other options would be to use Composer commands or to create a full-fledged login system.  Thoughts?

config/localConfig.multitenant.template.php
```diff
@@ -0,0 +1,94 @@
...
+
+/* Database Config */
+$db_type          = 'mysql'; // 'mysql' or 'pgsql'
+$db_host          = ''; // localhost or some other domain/ip
+$db_port          = '3306';
+$db_user          = '';
+$db_password      = '';
+$db_name          = '';
+$db_user_table    = 'users';
+$db_reports_table = 'reports';
```
bagofarms: Should add `$db_institutes_table` here.

lib/UdoitUtils.php
```diff
@@ -274,6 +274,15 @@ public function sortReportGroups($report_groups)
         return $ordered_report_groups;
     }
 
+    public function checkMultitenant($domain)
+    {
+        $result = UdoitDB::query('SELECT * FROM institutes WHERE domain = "'.$domain.'"')->fetchObject();
```
bagofarms: Use $db_institutes_table here instead of a constant to allow for installations where table names require prefixes or other special naming conventions.

multitenant/form_handler.php
webchuckweb: Agreed, but it feels like adding a full-fledged login system is overkill for managing a single table. Maybe Composer is a better option, but I have no experience in doing that.
Cidi Labs is actually using a separate system, Drupal, that will write to the institutes table in the customer setup process.

lib/UdoitUtils.php
iturgeon: Also suggest using PDO's prepared statements along with bindParam or bindValue to reduce sql injection vectors

see: https://github.com/ucfopen/UDOIT/blob/36a659a5693e3c26a34966ce75023642b57614a9/lib/UdoitUtils.php#L92-L94

multitenant/form_handler.php
bagofarms: Ahh, ok.  I didn't realize you were using a separate system.  I hate to lose the work you put into the admin interface, but I think it would be better to remove it from this pull request.  If an institution wants to do multitenant right now, they can add the institutions to the DB manually.  I will put in an issue for creating the Composer command to add institutions.

webchuckweb:
All good comments. I will try to work on these updates next week.

config/localConfig.template.php
```diff
@@ -47,21 +47,47 @@
...
+ * Valid values are 'ENV_PROD', 'ENV_DEV' (default), or 'ENV_TEST'
+ *
+ * ENV_DEV and ENV_TEST behaviors:
+ *  - set display errors to 1
+ *  - add test reports to 'View Old Reports' page in UDOIT
+ *  - No Canvas token validation
+ *  - Returns "test-token" when getting refresh token for Canvas
+ *  - Course Scanner doesn't scan, simply contains already loaded test reports
+ */
+$UDOIT_ENV = ENV_DEV; // default
```
bagofarms: Sorry I didn't notice this before, but it should really be `ENV_PROD` instead.  While in a perfect world it would be set to DEV for institutions to do some testing during implementation, I believe it would result in confusion.

bagofarms:
I accidentally merged this into 2.4.0 before it was ready.  I also don't think it will be a good fit for the 2.4.0 release.  Unfortunately, I can't reopen this pull request, so a new one will need to be opened after the following changes are made:

- The default `$UDOIT_ENV` should be `ENV_PROD`
- Any new code should have tests written for it
- The institutes table includes developer key information, but they don't need that since every institution should be using the global developer key.

PR #418, opened 2019-02-08
Add db_options to settings and pass to PDO object creation

dstepe:


bagofarms:
Thanks for the contribution!  A few bits of feedback:

Please change the base branch to `dev/v2.5.0` and resolve any merge conflicts.  Also, please add the same options for Postgres and Sqlite.

dstepe:
I rebased onto the 2.5.0 dev branch. I'm not so sure about adding options to the sqlite and pgsql types in the current implementation, though. From what I can tell, both will still use the PDO class, but neither uses the explicit username and password parameters. Since the options parameter is positioned after those two, they must be provided in the PDO class constructor in order to pass the options as well. Doing that would require more changes, and more risk, than I'm comfortable with given my current inexperience with the code.

Do you have suggestions on how to accomplish that without risking a breaking change?

bagofarms:
Looks good!  I'll test it out and merge it into 2.5.0.

Also, good point about Postgres and Sqlite.  I'll add the options in for Postgres since we're already modifying the PDO connection line in 2.5.0 to include the username and password explicitly.  I'll just leave sqlite the way it is since it's only used in our testing suite.

dstepe:
Sounds good. Thanks.

PR #429, opened 2019-06-26
Issue/377 open links in new tab

atarisafari:
Links in the preview section will now open in new tab


jrsilveti:
Confirmed working for both youtube and Vimeo links.

PR #430, opened 2019-06-27
Issue 412, changed error_logs to logger calls

atarisafari:


lib/Udoit.php
```diff
@@ -142,7 +143,7 @@ public static function scanContent(array $content_items)
 
                 $issue_count++;
                 $state = $quail_issue['state'];
-                //error_log("state in the issue is set as ".serialize($state));
+                $logger->addError("state in the issue is set as ".serialize($state));
```
bagofarms: Please remove this line.  It was inadvertently left in from some debugging, and error messages should only be logged when there's a program error, rather than Quail detecting an issue.

lib/quail/quail/quail.php
```diff
@@ -394,7 +395,7 @@ function runCheck($options = null)
 		}
 
 		$this->guideline = new $classname($this->dom, $this->css, $this->path, $options, $this->domain, $this->options['cms_mode']);
-		//error_log("runCheck completed, quail object is now: ".serialize($this));
+		$logger->addError("runCheck completed, quail object is now: ".serialize($this));
```
bagofarms: This one should be removed, as well.

PR #434, opened 2019-07-17
Issue/360 check caption language

### LINKED ISSUE #360: Check if video has captions based on course language
Currently videos are only checked to see if they have captions, not whether the captions are in the correct language. Perhaps look into cross-checking between course language or adding a new variable in the config.

atarisafari:
Fixes #360


atarisafari:
Fixes #360

config/localConfig.template.php
```diff
@@ -34,6 +34,9 @@
 /* Google Analytics Tracking Code */
 define('GA_TRACKING_CODE', '');
 
+/* Course Language */
+define('COURSE_LANGUAGE', '');
```
bagofarms: For consistency, please make the constant the same name as the environment variable in Heroku.

atarisafari:
Upgraded feature to not rely on config variables. Now just checks for the Canvas course language without any need for user setup.

atarisafari:
Note: Remind users in release notes that dialects will be treated as being the same language

bagofarms:
I did some testing and found that way more YouTube API requests were being made than expected.  I also talked to Matt about how we might be able to eliminate the extra API call for language by combining the two errors together and recognizing which one is which during the report building stage.

PR #438, opened 2019-07-18
Issue/431 unlisted videos should be suggestions

jrsilveti:
Merged Issue 200/ test for unavailable videos into dev/2-6-0.

Changed the errors for unlisted videos into suggestions.


jrsilveti:
Fixed unlisted videos showing up as an error and a suggestion.

bagofarms:
This got missed for 2.6.0 because it was missing the milestone.  Moved to 2.7.0

bagofarms:
Unfortunately, I'm going to have to close this issue without merging.  I tested the code, and it works perfectly, but the same feature was implemented in a different way as part of checking for caption language in #434 .  This only does a single API call rather than two, which is the primary reason for not merging this feature.

bagofarms:
Actually, I thought this over some more.  We have API caching within the same scan, so adding this shouldn't result in any extra API usage.  I will need to go in and remove the "manual check required" code from Ethan's code, though.

bagofarms:
Recreated in #547 so that I can make changes to the code.

PR #474, opened 2020-01-16
Issue/381 config file types

### LINKED ISSUE #381: Make "unscannable" filetypes configurable
Currently, UDOIT lists .pdf, .ppt, .pptx, .doc, and .docx files in "Unscannable".  We've received a request to add .xlsx files to that list, but maybe it would make more sense to have this be configurable?  This would affect [UDOIT.php](https://github.com/ucfopen/UDOIT/blob/882a5b9ac9cdd0768615bb99f27ad83fbdcf5efa/lib/Udoit.php#L249) and the filtering frontend in [default.js](https://github.com/ucfopen/UDOIT/blob/25cc9e0366b556fca9f5353c423fe92c1ab77af4/public/assets/js/default.js#L979)

atarisafari:
Fixes #381


bagofarms (approved):
Looks great!  Sorry it took so long.

PR #479, opened 2020-01-30
Issue/477 Reordering settings.php to allow for customized session params to be …

cidilabs:
…set in localConfig.php.

Bug fix: set default state of report summary button toggle to "collapse" rather than "expand".

### DIFF
--- a/config/settings.php
+++ b/config/settings.php
@@ -5,6 +5,14 @@ define('ENV_DEV', 'dev');
 
 define('UDOIT_VERSION', '2.6.0');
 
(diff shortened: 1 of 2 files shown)


cidilabs:
Fixed spacing for sniff test.

bagofarms:
Thank you again for submitting this.  We were having trouble with this working on Heroku due to how we assign the logger.  We have reordered things again in the #483 pr, and it seems to be working.  I'm going to close this pr, and ask for your feedback on the other one. @cidilabs

PR #480, opened 2020-01-30
Issue/478 safari iframe cookie fix

cidilabs:


bagofarms:
FYI, I reverted the commit that contains the settings reorder to isolate this change from the other one.

bagofarms:
The fix looks like it works, but it quickly refreshes the page and doesn't give any instructions after that.  If I manually refresh the whole page, UDOIT loads, but I get the following error at the top:  `Deprecated: Non-static method UdoitUtils::checkSafari() should not be called statically in /app/config/settings.php on line 59`

PR #483, opened 2020-02-06
Issue/477 cookiesettings fix

### LINKED ISSUE #477: SameSite cookie settings cannot be overwritten in localConfig.php
The current order of settings.php pulls in localConfig.php after the session cookie has been updated. With a little reordering we can pull in localConfig.php before we set the session cookie, so its parameters can be overwritten.

atarisafari:
Fixes #477


bagofarms:
This fix seems to work for Heroku, but I'm curious if it works properly for @webchuckweb

PR #488, opened 2020-02-13
Updating quail library to support newer HTML 5 spec for tables

### LINKED ISSUE #484: Update table checking code to include all variations of tables in HTML5
UFIXIT was used to fix the tables in a page, but the table is still flagged as missing headers.  This is the screenshot provided by Elisabeth Craft:

kennygperez:
Fixes #484


bagofarms:
Moving to 2.7.0 since it feels more like an enhancement than a bug.  It will also require more time to review.

bagofarms:
This seems to work really well!  While I was in here, I fixed #545

bagofarms (approved):
Thanks for contributing, Kenny!

PR #498, opened 2020-03-12
497 report ui filtering

cidilabs:
UI enhancements to the report. Adds tabs for content types and filters for issues. Most changes are done in a new JS and CSS file, very minimal changes to existing code.


cidilabs:
Not sure if I need to update this branch to be in line with 2.7.0, or if updating 2.7.0 to match the master branch will remove all the other commits that aren't mine.

bagofarms:
I updated this branch with Master, so only your changes should show up in this PR now.

Some feedback:  I noticed on your instance that there were two scrollbars.  Make sure you're calling `resizeFrame()` whenever the user does anything that will change the vertical height of the UI.

bagofarms:
Updated with the latest from 2.6.4 and 2.7.0 (also up to date with master).

During my desting, I discovered another concern.  The +/- button on "What does UDOIT look for?" also shows and hides the content selection boxes, but not the scan button that's right next to the content selection boxes.  I think moving the scan button below the content selection boxes would make this make more sense.  What do you think?

PR #565, opened 2021-04-22
Exit from a db error with a proper exit code

ottenhoff:
Also a few small changes in README

### DIFF
--- a/lib/UdoitDB.php
+++ b/lib/UdoitDB.php
@@ -133,7 +133,7 @@ class UdoitDB
             static::$last_test_time = time();
             static::$pdo = $db;
         } catch (\RuntimeException $e) {
-            $logger->addError("Database Connection Error: " . $e->getMessage());
+            $logger->addError("Database Connection Error: ".$e->getMessage());
             echo('Database Connection error');
             exit(1);
         }
--- a/lib/UdoitDB.php
+++ b/lib/UdoitDB.php
@@ -133,10 +133,9 @@ class UdoitDB
             static::$last_test_time = time();
             static::$pdo = $db;
         } catch (\RuntimeException $e) {
-            $logger->addError("Database Connection Error");
-            $logger->addError($e->getMessage());
+            $logger->addError("Database Connection Error: " . $e->getMessage());
             echo('Database Connection error');
-            exit();
+            exit(1);
         }
     }
 }
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ If you didn't already make `config/localConfig.php` when you set up the database
 Please refer to the [Canvas API Policy](http://www.canvaslms.com/policies/api-policy) before using this application, as it makes heavy use of the Canvas API.
 
 ### LTI Security
-UDOIT uses the security processes built into the LTI specification to ensure that users are only accessing UDOIT from within your instance of Canvas.  There are two values that need to be set in order for this security process to work.  These values should be different from each other.  You will use them again when you are installing the LTI in Canvas.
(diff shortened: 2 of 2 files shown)


lib/UdoitDB.php
```diff
@@ -135,8 +135,8 @@ protected static function connect()
         } catch (\RuntimeException $e) {
             $logger->addError("Database Connection Error");
             $logger->addError($e->getMessage());
-            echo('Database Connection error');
-            exit();
+            echo('Database Connection error: ' . $e->getMessage());
```
bagofarms, line 138: It's not useful to the user to know the exact error message, and the error message is already logged to `config/log.log` for the benefit of the developer/server admin.  Remove this change, and I will merge the rest of these changes.  We don't merge directly into master, so this will get merged into `dev/v2-8-0`

bagofarms (changes requested):
Remove the error message output from the echo statement.  See my comment on that line.

lib/UdoitDB.php
ottenhoff: dev/v2-8-0 looks like a tag ..... what's the general git strategy to follow?

bagofarms: Each release version has its own development branch.  Once that branch is merged into `master`, a tag is created with the version number.  For example, all of the bug fixes that released with v2.7.1 last night were originally merged into the `dev/v2-7-1` branch.  That let me do some final QA testing before release without touching the stable `master` branch yet.  When it was ready for release, I merged it into `master` and created a tag  called `v2.7.1` and [a release associated with it](https://github.com/ucfopen/UDOIT/releases/tag/v2.7.1).  You can read more in the [CONTRIBUTING.md](https://github.com/ucfopen/UDOIT/blob/v2.7.1/CONTRIBUTING.md#branching-and-merging) file.

ottenhoff: That's an excellent CONTRIBUTING doc! I do find it a bit unintuitive as a new developer. PRs from contributors are tough enough without having to know what is the correct devel branch. Is this a common Git workflow that I'm just not used to? Or is it possible to default PRs to a branch? 

My first instinct also is that it means features get lumped together which might have advantages and disadvantages. For example, what if you just wanted to remove one of the commits after an error was found in production (from master)? Would you roll back the entire merge from the branch? What is the major advantage instead of having each PR go into master? 

Sorry for all the questions!

bagofarms: No worries!  I can totally see how it's unintuitive at first.  Our setup is a modified version of [GitFlow](https://nvie.com/posts/a-successful-git-branching-model/).  We don't have a need for "nightly builds" of UDOIT, so we removed the `develop` branch entirely, and all feature branches (like `issue/539-httpful-version`) merge directly into release branches (like `dev/v2.7.1`).

You're right that features tend to get lumped together, but that usually works better for our user base. Most educational institutions we talked could not (or would rather not) update the application more than once a semester.  We also wanted to mitigate notification blindness by limiting the number of times we needed to post release notes on the Canvas Community.

If we find an error with a release that needs immediate attention, we can still create a `hotfix/some-problem` branch from `master`, fix the issue, merge it back into `master`, and create a patch-level version tag (according to SemVer) and release (which would be `v2.7.2` in this example).

ottenhoff: > could not (or would rather not) update the application more than once a semester

Yes, this is one of the biggest challenges for all higher-ed software. After years of doing very difficult once-per-year major upgrades, I would now try to do everything to move away from this old model and to move closer to an easy-updating, no-breaking, easy-rollback, release-often model. Staying on old software is bad for everyone.

bagofarms (approved):

Issue #394, opened 2018-12-17
'Avoid using styles for document structure' does not detect colored text that is bolded.


bagofarms:
This bug was discovered while validating #339 .  Currently, the 'Avoid using styles for document structure' suggestion will be thrown for 
`<p><strong><span style="color: #993300;">This text is inside a span</span></strong></p>`
but not for
`<p><span style="color: #993300;"><strong>This text is inside a strong</strong></span></p>`

PR #585, opened 2021-06-09
Issue/394 pNotUsedAsHeader ordering of strong and span

jasonmatos23-git:
Solves issue #394 by adding extra conditions to pNotUsedAsHeader test.


AlanFCMV (approved):
Confirmed it fixes the issue

PR #593, opened 2021-06-16
Issue/576 html markup encoding

### LINKED ISSUE #576: UFIXIT - HTMLEncoded HTML markup inside <code> block becomes unencoded after <img> alt text is fixed
An assignment has a ```<code>``` block containing HTML-Encoded HTML markup, and also an `````` tag that is missing alt text/decorative attribute:

```html
<pre>
    <code>
    &lt;strong&gt;weak&lt;/strong&gt;
    &lt;strong&gt;weaker&lt;/strong&gt;
    </code>
</pre>

<p>
    <img src="https://a.b.c/img.jpg" width="320" height="320" />
</p>
```

If UFIXIT is used to supply alt text / decorative marker, the HTML-Encoded markup becomes decoded and is then improperly rendered:

```html
<pre>
    <code>
        <strong>weak</strong>
        <strong>weaker</strong>
    </code>
</pre>

<p>
    <img src="https://a.b.c/img.jpg" alt="" width="320" height="320" data-decorative="true" />
</p>
```

jasonmatos23-git:
Closes #576 
This pr generally removes lines of code from Ufixit.php which encode/unencode the base html. I have tested using the code snippet provided in the issue and the fix has worked, but I would like to discuss this issue with @bagofarms before further review.


jasonmatos23-git:
Updates to Ufixit.php
- Removes &lt; and &gt; from annoying entities list (and < and > from replacements list). This was causing replacements to fail.
- Removes all instances where the context html is decoded/encoded. This was causing intentionally encoded characters to be decoded after submitting the ufixit.
Updates to process.php
- Removes decoding and special char filtering from error_html's definition. This was causing the error_html to become invalid html when a &lt; was present, which lead to errors in the correction.

AlanFCMV (approved):

PR #616, opened 2021-08-12
Add Github actions workflow yaml with linting

ottenhoff:


ottenhoff:
Are there any browser tests that should be run during CI? If not, is there a preferred solution for browser tests?

ottenhoff:
Hi @bagofarms, I know it's August (vacations + start of term), but what are the plans for CI? I definitely recommend moving to a PR-only workflow that requires each PR to pass tests, linting, etc. Direct pushes to main branches should be blocked via Github settings.

bagofarms:
@ottenhoff I'm 100% with you, and I'll be configuring the repo exactly like that once our first week of classes is over.

PR #639, opened 2021-09-03
Issue/638 Updates to migration and Heroku documentation

### LINKED ISSUE #638: Documentation updates for running migrations and Heroku
Instructions for executing migrations after running doctrine:migrations:diff is not correct. Also need to update some links within the HEROKU.md file.

jasonmatos23-git:
Closes #638

### DIFF
--- a/HEROKU.md
+++ b/HEROKU.md
@@ -35,7 +35,7 @@ heroku run bash
 ```
 Furthermore, you can access the Heroku Postgres database by running the following within that environment to execute SQL:
 ```
-psql <the DATABASE_URL from your config vars>
+psql <the DATABASE_URL located in your config vars under the Settings tab>
 ```
 Next we need to set up the database and insert our institution in to the appropriate table.
 1. Proceed with [setting up the database](https://github.com/ucfopen/UDOIT/blob/main/INSTALL.md#database-setup).
--- a/HEROKU.md
+++ b/HEROKU.md
@@ -30,32 +30,30 @@ git push heroku main:master -f
 
 ### Step 3: Database Migration and Setup
 Next we need to set up the database and insert our institution in to the appropriate table.
-1. Before setting up the database and if using Postgres, ensure that the table name in 'UDOIT/src/Entity/User.php' at line 14 is set to 'users'.
-  This change must be contained in a commit and be pushed to the Heroku remote before continuing.
-2. Access your Heroku instance by running
+1. Access your Heroku instance by running
 ```
 heroku run bash
 ```
-3. If using Heroku Postgres, generate a migration by using:
+2. If using Heroku Postgres, generate a migration by using:
 ```
 php bin/console doctrine:migrations:diff
 ```
-4. To execute the generated migration, run the following:
+3. To execute the generated migration, run the following:
 ```
 php bin/console doctrine:migrations:execute --up 'DoctrineMigrations\<Version################>'
 ```
   where '\<Version################\>' is the version that the previous command printed to the screen.
 
-5. If you are operating in a production environment you will need to generate the doctrine proxy classes by running the following command:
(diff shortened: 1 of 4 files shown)


HEROKU.md
```diff
@@ -35,16 +35,29 @@ heroku run bash
...
-psql <the DATABASE_URL from your config vars>
+psql <the DATABASE_URL located in your config vars under the Settings tab>
 ```
 Next we need to set up the database and insert our institution in to the appropriate table.
-1. Proceed with [setting up the database](https://github.com/ucfopen/UDOIT/blob/issue/570-heroku/INSTALL.md#database-setup).
-  Important Note: Before setting up the database and if using Postgres, ensure that the table name in 'UDOIT/src/Entity/User.php' at line 14 is set to 'users'.
-  This change must be contained in a commit and be pushed to the Heroku remote.
-2. Now is a good time to set up developer keys according to the instructions in INSTALL_\<LMS\>.md.
-3. Insert your institution in to the institution table as described in INSTALL_\<LMS\>.md.
+1. Before setting up the database and if using Postgres, ensure that the table name in 'UDOIT/src/Entity/User.php' at line 14 is set to 'users'.
```
bagofarms, line 41: Didn't we permanently change this table name to `users`?

HEROKU.md
```diff
@@ -35,16 +35,29 @@ heroku run bash
...
+  This change must be contained in a commit and be pushed to the Heroku remote before continuing.
+2. If using Heroku Postgres, generate a migration by using:
+```
+php bin/console doctrine:migrations:diff
```
bagofarms, line 45: Are these commands supposed to be run on the Heroku machine?  If so, add a step for running `heroku ps:exec`

bagofarms (changes requested):

HEROKU.md
jasonmatos23-git: We had discussed leaving it as 'user' so that cidilabs wouldn't have to change anything in their database. I can certainly change it to 'users' though, that would make things simpler for us on Heroku

HEROKU.md
jasonmatos23-git: I moved the line for using `heroku run bash` into a step under the 'Database Migration and Setup' section. I decided not to go with `heroku ps:exec` based on this [post about using bash for database migrations](https://stackoverflow.com/questions/54386370/heroku-what-is-the-difference-between-heroku-psexec-and-heroku-run-bash).

HEROKU.md
jasonmatos23-git: I removed the step that mentions modifying this table name in anticipation for permanently changing it to `users`. This PR has the table name set as `users`, but I assume we should make a separate one for that change?

bagofarms: Yes, please make a separate PR that changes that table name and includes a migration so that existing installations can update their databases easily.

jasonmatos23-git: #647 will contain the table name modification and migration

HEROKU.md
```diff
@@ -29,22 +29,33 @@ git push heroku main:master -f
...
+Next we need to set up the database and insert our institution in to the appropriate table.
+1. Access your Heroku instance by running
 ```
 heroku run bash
 ```
-Furthermore, you can access the Heroku Postgres database by running the following within that environment to execute SQL:
+2. If using Heroku Postgres, generate a migration by using:
 ```
-psql <the DATABASE_URL from your config vars>
+php bin/console doctrine:migrations:diff
```
bagofarms, line 39: You probably need to remove this since the postgres migration stuff is in a separate PR

bagofarms (approved):
I'm still a little confused about how the self-hosted install instructions will work around the heroku migration, but I'm willing to merge this for now.  I will try setting up my local Docker environment and put in a new issue if I have problems

PR #647, opened 2021-09-09
Issue/646 Change User entity table name from user to users

### LINKED ISSUE #646: Change User entity table name to 'users'
Change the User entity table name from 'user' to 'users' in order to support Postgres.

jasonmatos23-git:
Closes #646

### DIFF
--- a/src/DoctrineMigrations/Version20210910203121.php
+++ b/src/DoctrineMigrations/Version20210910203121.php
@@ -20,40 +20,12 @@ final class Version20210910203121 extends AbstractMigration
     public function up(Schema $schema): void
     {
         // this up() migration is auto-generated, please modify it to your needs
-        $this->addSql('ALTER TABLE file_item DROP FOREIGN KEY FK_96E74D70FC6B21F1');
-        $this->addSql('ALTER TABLE issue DROP FOREIGN KEY FK_12AD233EC38008F2');
-        $this->addSql('ALTER TABLE log_entry DROP FOREIGN KEY FK_B5F762DA76ED395');
-        $this->addSql('ALTER TABLE report DROP FOREIGN KEY FK_C42F7784F675F31B');
-        $this->addSql('CREATE TABLE users (id INT AUTO_INCREMENT NOT NULL, institution_id INT NOT NULL, username VARCHAR(180) NOT NULL, name VARCHAR(255) DEFAULT NULL, roles JSON DEFAULT NULL, lms_user_id VARCHAR(128) NOT NULL, api_key VARCHAR(2048) DEFAULT NULL, refresh_token VARCHAR(512) DEFAULT NULL, created DATETIME NOT NULL, last_login DATETIME NOT NULL, UNIQUE INDEX UNIQ_1483A5E9F85E0677 (username), INDEX IDX_1483A5E910405986 (institution_id), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_unicode_ci` ENGINE = InnoDB');
-        $this->addSql('ALTER TABLE users ADD CONSTRAINT FK_1483A5E910405986 FOREIGN KEY (institution_id) REFERENCES institution (id)');
-        $this->addSql('DROP TABLE user');
-        $this->addSql('ALTER TABLE file_item DROP FOREIGN KEY FK_96E74D70FC6B21F1');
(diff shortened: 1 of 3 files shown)


src/DoctrineMigrations/Version20210910203121.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE file_item DROP FOREIGN KEY FK_96E74D70FC6B21F1');
+        $this->addSql('ALTER TABLE issue DROP FOREIGN KEY FK_12AD233EC38008F2');
+        $this->addSql('ALTER TABLE log_entry DROP FOREIGN KEY FK_B5F762DA76ED395');
+        $this->addSql('ALTER TABLE report DROP FOREIGN KEY FK_C42F7784F675F31B');
+        $this->addSql('CREATE TABLE users (id INT AUTO_INCREMENT NOT NULL, institution_id INT NOT NULL, username VARCHAR(180) NOT NULL, name VARCHAR(255) DEFAULT NULL, roles JSON DEFAULT NULL, lms_user_id VARCHAR(128) NOT NULL, api_key VARCHAR(2048) DEFAULT NULL, refresh_token VARCHAR(512) DEFAULT NULL, created DATETIME NOT NULL, last_login DATETIME NOT NULL, UNIQUE INDEX UNIQ_1483A5E9F85E0677 (username), INDEX IDX_1483A5E910405986 (institution_id), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_unicode_ci` ENGINE = InnoDB');
+        $this->addSql('ALTER TABLE users ADD CONSTRAINT FK_1483A5E910405986 FOREIGN KEY (institution_id) REFERENCES institution (id)');
+        $this->addSql('DROP TABLE user');
```
bagofarms, line 29: Can we alter the existing table name instead of creating a new one and dropping the old one?  I don't want existing installs to lose all of their user data.

src/DoctrineMigrations/Version20210910210740.php
```diff
@@ -0,0 +1,113 @@
...
+        $this->addSql('CREATE UNIQUE INDEX UNIQ_1483A5E9F85E0677 ON users (username)');
+        $this->addSql('CREATE INDEX IDX_1483A5E910405986 ON users (institution_id)');
+        $this->addSql('CREATE TABLE messenger_messages (id BIGSERIAL NOT NULL, body TEXT NOT NULL, headers TEXT NOT NULL, queue_name VARCHAR(190) NOT NULL, created_at TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL, available_at TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL, delivered_at TIMESTAMP(0) WITHOUT TIME ZONE DEFAULT NULL, PRIMARY KEY(id))');
+        $this->addSql('CREATE INDEX IDX_75EA56E0FB7336F0 ON messenger_messages (queue_name)');
+        $this->addSql('CREATE INDEX IDX_75EA56E0E3BD61CE ON messenger_messages (available_at)');
+        $this->addSql('CREATE INDEX IDX_75EA56E016BA31DB ON messenger_messages (delivered_at)');
+        $this->addSql('CREATE OR REPLACE FUNCTION notify_messenger_messages() RETURNS TRIGGER AS $$
```
bagofarms, line 57: Postgres really needs all of these changes in order to work?

src/DoctrineMigrations/Version20210910203121.php
jasonmatos23-git: Replaced all of that with a simple rename command

src/DoctrineMigrations/Version20210910210740.php
jasonmatos23-git: Yes Postgres does need these to work. The code could be condensed with psuedo-classes (e.g., `SERIAL` for the ID incrementing) though, if that sounds desirable.

jasonmatos23-git:
I have tested the two newly added migration's `up` and `down` functions, and both work as intended.

bagofarms (approved):
Thanks!

ottenhoff:
Is MySQL/Maria still supported? If yes, this isn't going to work:

`updated TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL,`

`
In PDOConnection.php line 132:

  SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'TIME
  ZONE NOT NULL, metadata TEXT DEFAULT NULL, published BOOLEAN NOT NULL, a...' at line 1

`

PR #650, opened 2021-09-14
Github Actions with Database Creation

ottenhoff:
This is currently failing, but the idea is to test all supported DBs during the PR CI run.

### DIFF
--- a/.github/workflows/udoit.yml
+++ b/.github/workflows/udoit.yml
@@ -21,7 +21,7 @@ jobs:
         uses: shivammathur/setup-php@v2 #https://github.com/shivammathur/setup-php
         with:
           php-version: ${{ matrix.php-versions }}
-          extensions: mbstring, xml, ctype, iconv, intl, pdo_sqlite, gd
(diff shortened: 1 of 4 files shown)


ottenhoff:
I'm assuming that these doctrine migrate runs should work on both Postgres and MySQL on a fresh install?

bagofarms:
@jasonmatos23-git Should those migrations work on MySQL?

jasonmatos23-git:
The only migrations that work on MySQL are Version20210910203121 which is to rename the 'user' table to 'users', and Version20210525163808 which is the original migration provided by cidilabs. Neither of them works with Postgres.
Version20210910210740 is the only one that works for Postgres, and does not work with MySQL.

There also does not seem to be a way to combine the two database setup migrations because of differences in function names and data types, for example 'timestamp without time zone' is 'datetime' in MySQL

ottenhoff:
Thanks but now I'm even more confused as to DB support in main branch. How does a new user stand up a database?

bagofarms:
@ottenhoff It looks like you got it right.  I'm going to test this right now and hopefully get it merged today.

bagofarms (approved):

Issue #643, opened 2021-09-07
Colored text inside H2 flagged for suggestion, UFIXIT not fixing


bagofarms:
In UDOIT 2.8.2, an H2 containing colored text is flagged for the "Avoid using color alone for emphasis" suggestion.  The typical use case that would generate this HTML is a teacher trying to style their course by applying color to their headings.  I don't think this should be flagged as a suggestion in that case.  However, a teacher may try to emphasize a single word in a heading by applying a color to it.  In that case, the suggestion is valid.

Additionally, running UFIXIT on this HTML does not fix the issue.  The HTML in the page appears unaffected.

Example HTML:
```html
<h2><span style="color: #993300;">This is a heading with color applied</span></h2>
```

This issue is for 2.x, but I would like someone to see if it exists in 3.x.  If so, please fix!

AlanFCMV:
I can confirm that this is still a problem in 3.x. The UFIXIT for this and the color contrast rule is also broken, it completely removes the color and background color from the span.
Before hitting save:
![image](https://user-images.githubusercontent.com/54645815/132518245-5368647f-73ee-4ac6-acce-7377974393f0.png)
`<h2><span style="color: #993300;">This is a heading with color applied</span></h2>
`
![image](https://user-images.githubusercontent.com/54645815/132518472-87a35e8f-2b58-4446-9a02-b04737ae980f.png)
After hitting save:
![image](https://user-images.githubusercontent.com/54645815/132518674-ecf8d9a1-e6fe-4def-abb2-b168855d9fca.png)
`<h2><span style="font-weight: normal; font-style: normal;">This is a heading with color applied</span></h2>`

AlanFCMV:
This same issue happens with the background color contrast ufixit.

![image](https://user-images.githubusercontent.com/54645815/132520448-5d9855ab-6dc0-4123-baad-0452d7de3d28.png)

After pressing save:

![image](https://user-images.githubusercontent.com/54645815/132520571-abe1333d-4a59-488e-8f5a-47eb74da2321.png)
`<p><span style="font-weight: normal; font-style: normal;">Here is another bit of text.</span></p>
`

AlanFCMV:
![image](https://user-images.githubusercontent.com/54645815/135291274-9bd234d5-905b-4a48-8960-ff09b95cd76e.png)
The default text color isn't black.

PR #651, opened 2021-09-15
Fix color contrast ufixit

AlanFCMV:
Closes issue #643

### DIFF
--- a/assets/js/Services/Contrast.js
+++ b/assets/js/Services/Contrast.js
@@ -3,11 +3,11 @@ class Contrast {
     if (/^#[0-9A-F]{6}$/i.test(rgb)) return rgb;
 
     rgb = rgb.match(/^rgb\((\d+),\s*(\d+),\s*(\d+)\)$/);
-    
+
     if (!rgb) {
       return rgb;
     }
-    
+
     function hex(x) {
       return ("0" + parseInt(x).toString(16)).slice(-2);
     }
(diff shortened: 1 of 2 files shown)


assets/js/Services/Contrast.js
```diff
@@ -225,6 +225,14 @@ class Contrast {
     return '#' + color
   }
 
+  convertHtmlRgb2Hex(html) {
+    return html.replace(/rgb\((.+?)\)/ig, (_, rgb) => {
+      return '#' + rgb.split(',')
+        .map(str => parseInt(str, 10).toString(16).padStart(2, '0'))
+        .join('')
+    })
+  }
```
bagofarms, line 234: This will also replace all occurrences of rgb color values in the text of the page, which could be bad for any classes that are teaching CSS or HTML.  I suggest something like this, which would target only rgb color values that are within an HTML tag:
```js
convertHtmlRgb2Hex(html) {
  return html.replace(/(<[a-z"=;#:\-\d\s]*rgb\()(.+?)(\)+[a-z"=;#:\-\s]*>)/ig, (_, pre, rgb, post) => {
    return pre + '#' + rgb.split(',')
      .map(str => parseInt(str, 10).toString(16).padStart(2, '0'))
      .join('') + post
  })
}
```

You can try it out here:  https://codepen.io/bagofarms/pen/ExXoOXe

bagofarms (changes requested):

assets/js/Services/Contrast.js
AlanFCMV: That's a great suggestion. That specific implementation won't quite work when a tag has multiple attributes with an rgb value such as color and background-color, and the parentheses are kept around. 

I went ahead and fixed those issues and implemented the change to ignore rgb elements that aren't inside html tags.

bagofarms (approved):

Issue #697, opened 2021-11-03
Remove "Embed must have a noembed rule" from UDOIT


cidilabs:
The rule "EmbedHasAssociatedNoEmbed" is based on deprecated HTML and needs to be removed.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/noembed

Noembed is a non-standard feature and should not be encouraged.  

Also, Canvas no longer supports <noembed>. If you manually enter it, the RCE will strip it out.

I have removed the rule from PhpAlly. This ticket is a reminder that we need to remove it from the front-end as well.

PR #698, opened 2021-11-04
removed deleted rules from constants.js

atarisafari:
Addresses #697


AlanFCMV (changes requested):
The deleted rules should also be removed from the .env file. To be specific, ObjectInterfaceIsAccessible and ObjectShouldHaveLongDescription are both currently listed under PhpAlly suggestions.

Additionally, text associated with these rules should be deleted from en.json and es.json.

AlanFCMV (approved):

PR #702, opened 2021-11-05
Hopeful fix for bug that sets activeIssue incorrectly after issue save.

cidilabs:


cidilabs:
@AlanFCMV This is actually a "feature" rather than a bug. When we originally developed it the current issue would immediately disappear from the modal once you saved or marked as resolved, and the user experience was very jarring. So we added this intermediate state that keeps the resolved issues until the list is refreshed. 

If you feel that this is confusing to the user let's create an issue to discuss other alternatives.

AlanFCMV (approved):

PR #729, opened 2021-12-13
Fixes #727 by looking for LTI initiation in both POST and GET per IMS specification

### LINKED ISSUE #727: LTI 1.3 initiation from LMS can be GET or POST
Current code in src/Controller/LtiController.php only looks for POST parameters. I believe this can be both GET and POST:

https://www.imsglobal.org/forum/learning-tools-interoperability-public-forum/196841

ottenhoff:


src/Controller/LtiController.php
```diff
@@ -263,10 +263,19 @@ protected function saveTokenToSession($token)
     protected function saveRequestToSession()
     {
         try {
+            $getParams = $this->request->query->all();
             $postParams = $this->request->request->all();
```
rob-3, line 267: @ottenhoff What do you think of shortening this by combining the arrays and then doing just a single loop? I.E. something like
```diff
         try {
             $getParams = $this->request->query->all();
             $postParams = $this->request->request->all();
+            $allParams = array_merge($getParams, $postParams);
 
-            foreach ($getParams as $key => $val) {
-                if (!empty($val)) {
-                    $this->session->set($key, $val);
-                }
-            }
-
-            foreach ($postParams as $key => $val) {
+            foreach ($allParams as $key => $val) {
                 if (!empty($val)) {
                     $this->session->set($key, $val);
                 }
```

rob-3 (changes requested):

src/Controller/LtiController.php
ottenhoff: Yeah, I considered just combining the two arrays, but I decided that would make debugging more difficult in case of an issue where URL params were being sent in addition to POST params. I thought this way was more explicit in demonstrating that POST params would take precedence.

ottenhoff: Okay, array_merge implemented.

rob-3 (approved):

PR #730, opened 2021-12-13
Make the LTI JWKS exchange more general for all LMSes

ottenhoff:
The current LTI exchange is very Canvas-specific. 

The LTI spec (http://www.imsglobal.org/spec/lti/v1p3/#tool-deployment) details:

* sub: required user identity info
* given_name: optional
* family_name: optional
* email: optional
* context: optional contains info about where tool launched from

### DIFF
--- a/src/Controller/LtiController.php
+++ b/src/Controller/LtiController.php
@@ -101,7 +101,11 @@ class LtiController extends AbstractController
 
         if (isset($token->{'https://purl.imsglobal.org/spec/lti/claim/target_link_uri'})) {
             $redirectUrl = $token->{'https://purl.imsglobal.org/spec/lti/claim/target_link_uri'} . '?auth_token=' . $this->session->getUuid();
-            return $this->redirect($redirectUrl);
+
+            // Avoid a never-ending redirect loop
+            if (strpos($redirectUrl, $this->request->getRequestUri()) === false) {
+                return $this->redirect($redirectUrl);
+            }
         }
 
         return $this->redirectToRoute('dashboard',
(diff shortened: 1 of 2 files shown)


bagofarms:
Thank you for this PR.  We're back from the holiday break, and we'll start reviewing this (and your other PRs) soon.

src/Controller/LtiController.php
```diff
@@ -239,6 +243,14 @@ protected function saveTokenToSession($token)
...
             }
 
+            // Context is optional and contains a unique id, label, title, and type
+            if (!empty($token->{'https://purl.imsglobal.org/spec/lti/claim/context'})) {
+                $contextFields = (array) $token->{'https://purl.imsglobal.org/spec/lti/claim/context'};
+                foreach ($contextFields as $key => $val) {
+                    $this->session->set('lms_course_' . $key, $val);
+                }
+            }
+
```
rob-3, line 253: The relocation of this code from `src/Lms/D2l/D2lLms.php` seems to have broken UDOIT for us. Any reason why this was moved?

rob-3: @ottenhoff

rob-3: Strangely, it causes our Canvas instance to get an API error

rob-3: I'm going to revert and reopen this until I resolve the issue

ottenhoff: Maybe you're getting LMS course information from two places: the LTI way and the custom-Canvas-API way?

If the goal is to get this project to be less-custom-Canvas and more-LTI-general, then trying to do things the proper-LTI way makes sense. The claim context is the standard way to get info on the launching course

rob-3: I 100% agree on this. I'll do some digging tomorrow and see if I can't figure it out. UCF uses canvas so the majority of our testing is there, but the standard protocols should be properly supported.

PR #737, opened 2022-01-12
Revert "Make the LTI JWKS exchange more general for all LMSes"

rob-3:
Reverts ucfopen/UDOIT#730. This caused an issue in our Canvas installation, but will be remerged when we fix that.

### DIFF
--- a/src/Controller/LtiController.php
+++ b/src/Controller/LtiController.php
@@ -101,11 +101,7 @@ class LtiController extends AbstractController
 
         if (isset($token->{'https://purl.imsglobal.org/spec/lti/claim/target_link_uri'})) {
             $redirectUrl = $token->{'https://purl.imsglobal.org/spec/lti/claim/target_link_uri'} . '?auth_token=' . $this->session->getUuid();
-
-            // Avoid a never-ending redirect loop
-            if (strpos($redirectUrl, $this->request->getRequestUri()) === false) {
-                return $this->redirect($redirectUrl);
-            }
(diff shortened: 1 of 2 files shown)


bagofarms:
Rob will continue working on the real solution to this problem tomorrow, but for now let's revert 3.2.0 to "fix" the issue.

ottenhoff:
Reverting from a community branch seems like a strange workflow unless you know that the commit is wrong/incorrect?

rob-3:
This branch is the development branch; we didn't and no one should use this branch for production. While this PR looks to be good for the project, it's my fault for marking it ready to merge without thoroughly testing it. I caught the error while testing an unrelated change on a development Canvas deployment.

I think that the breakage in our Canvas deployment should at least be investigated to ensure that the issue is entirely on our side prior to merging this in. I don't want to trade an existing issue for a new one.

I apologize for any inconvenience this causes you; I am a new member on the team and am still getting up to speed with the proper processes for maintaining an open source project like this.

ottenhoff:
It's not a big deal, and it's a young project still with room to change processes.

In general, making it difficult to get a PR in makes sense. This should be done with lots of automated tests, integration tests, browser tests, unit tests, and manual PR review.

But as a PR is difficult to merge in, it should have a difficult bar to remove. If a PR generally moves things in the right direction, fixing it instead of reverting usually makes more sense. But reversion makes sense sometimes, especially when a PR ends up complicating things and causing many regressions.

PR #738, opened 2022-01-13
Redirect logs to stderr

rob-3:
This should make actually viewing logs in Docker and Heroku possible.


rob-3:
Closed in favor of #741, which doesn't include unrelated changes.

PR #744, opened 2022-01-19
Use $util->getCurrentDomain() instead of JWK_BASE_URL

### LINKED ISSUE #740: Unable to use hardcoded value of JWK_BASE_URL for beta/test installs and multiple instituions
In testing this, it seems like there is only one hardcoded environment variable `JWK_BASE_URL` for all hosted institutions to use in https://github.com/ucfopen/UDOIT/blob/d19ef66a7fd70d45f85d17b1ba1ec6e4c3a64f65/src/Lms/Canvas/CanvasLms.php#L79

The value by default goes to https://canvas.instructure.com/api/lti/security/jwks and this seems to match the keys generated by the **production** Canvas installation for launch.

However in the **beta** and **test** instances, it looks to require use of the domain which is

`$baseURL = $this->util->getCurrentDomain();`

This would go to this, similar to other calls
`https://${baseURL}/api/lti/security/jwks`

I think we'd either need to

*  Always use this as the `$baseURL` for all instances, it works fine for production and you don't need to go to the generic canvas.instructure.com URL
* Use a specific `JWK_URL` for each institition, defined in the database institution table.

rob-3:
This is intended to resolve #740

### DIFF
--- a/HEROKU.md
+++ b/HEROKU.md
@@ -18,7 +18,8 @@ Click the Heroku button and follow the instructions below:
    * `d2l` if you are using the D2l Brightspace LMS.
 5. Fill out the `BASE_URL` field with `https://yourapp.herokuapp.com`. (Replace 'yourapp' with the name you gave in step 1.2.)
 6. Fill out the `JWK_BASE_URL` field with the URL to your LMS. The default value works for instructure hosted instances of Canvas, but will need to be modified according to the LMS and host.
-7. Click the Deploy button and wait for the process to complete.
+7. Fill out the `LMS_BASE_DOMAIN` field with a bare domain (no "https://" or trailing slash) of the LMS.
+8. Click the Deploy button and wait for the process to complete.
 
 ### Step 2: Database Migration and Setup
 Next we need to set up the database and insert our institution in to the appropriate table.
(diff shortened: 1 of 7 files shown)


bagofarms:
It looks like this won't work for the use case in #740, since they have one instance of UDOIT and would like to use it on their prod and test/beta instances of Canvas at the same time.  This is one setting that all institutions must share.

rob-3:
> It looks like this won't work for the use case in #740, since they have one instance of UDOIT and would like to use it on their prod and test/beta instances of Canvas at the same time. This is one setting that all institutions must share.

Oh I see the issue now. The previous approach read the domain from the `iss` session variable, but our test canvas instance sets that to `canvas.instructure.com`, which caused it to break. I was told this was a valid configuration, but perhaps we can have a variable that contains the actual domain.

cidilabs:
I think this PR was on the right track, then took a wrong turn. In most cases the `iss` value is the correct domain for the LMS instance. 

Here is one update we've made that somehow didn't get pushed back to UCFOPEN.
CanvasLms.php
```
    public function getKeysetUrl()
    {
        $session = $this->sessionService->getSession();
        $baseUrl = $session->get('iss');

        return "{$baseUrl}/api/lti/security/jwks";
    }
```

rob-3:
The problem is that in our setup, iss doesn't point to the right url and I was under the impression that iss isn't guaranteed to point to what we need.

rob-3:
Closing in favor of #747, which should resolve this by reading from `JWK_BASE_URL` if it exists but falling back to reading `iss` if not

PR #747, opened 2022-01-25
CanvasLms updates that didn't get pushed to UCFOPEN

cidilabs:
CanvasLms updates that didn't get pushed to UCFOPEN:
- use $session->get('iss') instead of $_ENV['JWK_BASE_URL']
- store terms in session for admin page
- typo in quiz option, createLmsPostOptions()
- track module items with external URLs

### DIFF
--- a/src/Lms/Canvas/CanvasLms.php
+++ b/src/Lms/Canvas/CanvasLms.php
@@ -87,6 +87,10 @@ class CanvasLms implements LmsInterface {
 
     public function getKeysetUrl()
     {
+        if ($_ENV['JWK_BASE_URL']) {
+            $jwk_url_no_trailing_slash = rtrim($_ENV['JWK_BASE_URL'], '/');
+            return "{$jwk_url_no_trailing_slash}/api/lti/security/jwks";
+        }
         $session = $this->sessionService->getSession();
         $baseUrl = $session->get('iss');
 
--- a/src/Lms/Canvas/CanvasLms.php
+++ b/src/Lms/Canvas/CanvasLms.php
@@ -641,11 +641,11 @@ class CanvasLms implements LmsInterface {
 
             case 'assignment':
                 $out['id'] = $lmsContent['id'];
-                $out['url'] = "{$baseUrl}/assignments/{$lmsContent['id']}";
                 $out['title'] = $lmsContent['name'];
                 $out['updated'] = $lmsContent['updated_at'];
                 $out['body'] = $lmsContent['description'];
                 $out['status'] = $lmsContent['published'];
+                $out['url'] = "{$baseUrl}/assignments/{$lmsContent['id']}";
 
                 break;
 
--- a/src/Lms/Canvas/CanvasLms.php
+++ b/src/Lms/Canvas/CanvasLms.php
@@ -73,14 +73,12 @@ class CanvasLms implements LmsInterface {
      * ********************
      */
 
-    public function getLtiAuthUrl($globalParams)
+    public function getLtiAuthUrl($params)
     {
         $session = $this->sessionService->getSession();
         $baseUrl = $_ENV['JWK_BASE_URL'] ?: $session->get('iss');
         $baseUrl = rtrim($baseUrl, '/');
 
-        $lmsParams = [];
-        $params = array_merge($globalParams, $lmsParams);
         $queryStr = http_build_query($params);
 
         return "{$baseUrl}/api/lti/authorize_redirect?{$queryStr}";
--- a/src/Lms/Canvas/CanvasLms.php
+++ b/src/Lms/Canvas/CanvasLms.php
@@ -88,12 +88,9 @@ class CanvasLms implements LmsInterface {
 
     public function getKeysetUrl()
     {
-        if ($_ENV['JWK_BASE_URL']) {
-            $jwk_url_no_trailing_slash = rtrim($_ENV['JWK_BASE_URL'], '/');
-            return "{$jwk_url_no_trailing_slash}/api/lti/security/jwks";
-        }
         $session = $this->sessionService->getSession();
-        $baseUrl = $session->get('iss');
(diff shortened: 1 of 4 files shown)


rob-3:
My only concern is that the spec doesn't require `iss` to be the domain of the keyset URL. I'm not sure if this is an issue in practice, but it's technically non-compliant behavior.

Even if `iss` is `mysetup.edu`, the keyset URL could be `security.mysetup.edu/jwks` or `someotherdomain.edu/jwks`, which breaks this.

cidilabs:
That's a valid point. But for Canvas I believe the `iss` will always be a domain that points to a Canvas instance, so `<canvas_url>/api/lti/security/jwks` should work. And for any other LMS, they can define a different URL in their LMS class.

rob-3:
We may have a misconfigured instance then, because our `iss` for our self hosted canvas instance is `canvas.instructure.com`.

cidilabs:
Right, that will be the case for most Canvas institutions in prod. Test and beta instances will be https://canvas.test.instructure.com and https://canvas.beta.instructure.com. These are the URLs we use for authentication. See step 1: https://canvas.instructure.com/doc/api/file.lti_dev_key_config.html

rob-3:
Sorry, I said that in a confusing way. What I meant to say is that our *development* canvas instance, which is self-hosted at another domain still has that `iss: canvas.instructure.com`. Perhaps that is a misconfiguration.

src/Lms/Canvas/CanvasLms.php
```diff
@@ -51,7 +51,7 @@ public function __construct(
         UtilityService $util,
         Security $security,
         SessionService $sessionService,
-        HtmlService $html)
+        HtmlService $html)
```
rob-3, line 54: These trailing spaces are just creating noise in the diff; would you mind removing them?

src/Lms/Canvas/CanvasLms.php
```diff
@@ -631,11 +641,11 @@ protected function normalizeLmsContent(Course $course, $contentType, $lmsContent
 
             case 'assignment':
                 $out['id'] = $lmsContent['id'];
+                $out['url'] = "{$baseUrl}/assignments/{$lmsContent['id']}";
                 $out['title'] = $lmsContent['name'];
                 $out['updated'] = $lmsContent['updated_at'];
                 $out['body'] = $lmsContent['description'];
                 $out['status'] = $lmsContent['published'];
-                $out['url'] = "{$baseUrl}/assignments/{$lmsContent['id']}";
```
rob-3, line 638: Is there a reason this line of code moved?

rob-3 (changes requested):
This mostly lgtm; just a couple of things in the diff that seemed strange to me.

rob-3:
We also decided that we like the `iss` change, but we would like to have UDOIT override `iss` with `JWK_BASE_URL` if it exists to accommodate less conventional setups.

rob-3:
@cidilabs, I have made a few changes to this PR; lmk if you're still happy to merge this in.

mrash2:
Some of this is above my head. Would any of this cause the message "Session authentication failed" to appear upon launching UDOIT from a Canvas course? That's as far as we're able to get with our self-hosted UDOIT install + our hosted test Canvas instance. We're using the 3.0 release from UCFOPEN.

rob-3:
This sounds like a different issue; if you file an issue in the issue tracker with more details, I can look into this.

src/Lms/Canvas/CanvasLms.php
```diff
@@ -76,7 +76,8 @@ public function getId()
     public function getLtiAuthUrl($globalParams)
     {
         $session = $this->sessionService->getSession();
-        $baseUrl = $session->get('iss');
+        $baseUrl = $_ENV['JWK_BASE_URL'] ?: $session->get('iss');
```
cidilabs, line 79: This will return a notice if `JWK_BASE_URL` is undefined in .env. Since the variable is optional let's try:
```
$baseUrl = !empty($_ENV['JWK_BASE_URL']) ? $_ENV['JWK_BASE_URL'] : $session->get('iss');
```
This will ignore `JWK_BASE_URL` if it's an empty string, null, or undefined.

src/Lms/Canvas/CanvasLms.php
```diff
@@ -88,12 +88,9 @@ public function getLtiAuthUrl($globalParams)
 
     public function getKeysetUrl()
     {
-        if ($_ENV['JWK_BASE_URL']) {
-            $jwk_url_no_trailing_slash = rtrim($_ENV['JWK_BASE_URL'], '/');
-            return "{$jwk_url_no_trailing_slash}/api/lti/security/jwks";
-        }
         $session = $this->sessionService->getSession();
-        $baseUrl = $session->get('iss');
+        $baseUrl = $_ENV['JWK_BASE_URL'] ?: $session->get('iss');
```
cidilabs, line 90: Same as mentioned before. I love the elvis operator, but in this case I think it's better to use `!empty()` since `JWK_BASE_URL` is an optional parameter.

cidilabs:
@rob-3 Thanks for adding these changes. I made a few minor comments, but overall looks great.

src/Lms/Canvas/CanvasLms.php
rob-3: I agree with this suggestion; I'm also switching to `getenv()` as that appears to be the modern best practice.

src/Lms/Canvas/CanvasLms.php
```diff
@@ -87,6 +87,10 @@ public function getLtiAuthUrl($globalParams)
 
     public function getKeysetUrl()
     {
+        if ($_ENV['JWK_BASE_URL']) {
```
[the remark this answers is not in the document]
rob-3: I will switch to `getenv()`, which doesn't cause any warnings and returns false if the variable is not defined.

rob-3:
@cidilabs if you can give this one more look, I will merge it in.

PR #758, opened 2022-02-02
Fix auto scripts warning

rob-3:
When you run `composer install`, you get a warning about auto-scripts. We aren't using that feature anyway, so this removes it to fix the warning.


SimonRothUCF:
No errors here, but it's not liking that `composer.json` file for merging.

rob-3:
Looks like I already fixed this in b7364895 and just forgot about it.

PR #759, opened 2022-02-02
Fix webpack public path

rob-3:


SimonRothUCF:
Looks good to me, merging

PR #761, opened 2022-02-03
Upgrade to Symfony 5.4

rob-3:
~This probably is going to go into 3.3.0; feel free to change the base~ done


rob-3:
Looks like this will force us to drop support for PHP 7.4 when it goes in, so we might want to do a major version bump to 4.0.0.

PR #766, opened 2022-02-07
Add nonce validation to SessionService

rob-3:
@bagofarms I would appreciate if you could review this as it is a security matter and I want to make sure it is correct. I'm still planning to update the code to return a non 500 error message but the security stuff should be done.

~Edit: something seems to be broken already; let me investigate~
Edit: ready to merge whenever this can be reviewed


src/Services/NonceService.php
```diff
@@ -0,0 +1,35 @@
...
+    public function __construct(SessionService $sessionService)
+    {
+        $this->sessionService = $sessionService;
+    }
+
+    public function generateNonce(string $uuid): string {
+        $session = $this->sessionService->getSession($uuid);
+        $nonce = bin2hex(random_bytes(8));
+        $session->set('nonce', [$nonce, time()]);
+        $session->set('hello', 'abc123');
```
bagofarms, line 19: What is this for?

rob-3: Oh that's a mistake. It must've gotten left from when I was debugging the issues

rob-3:
Taking a look at this again; I'm wondering if it makes more sense to just integrate this into the `SessionService`. I could just move the methods and then we wouldn't need a separate service. What do you think @bagofarms?

bagofarms:
@rob-3 If it's only ever used in relation to a session, I think it would make more sense integrated into the `SessionService`.

PR #779, opened 2022-02-10
Fix `User` and `UserInterface` intelphense errors

### LINKED ISSUE #775: `User` and `UserInterface` interfaces are incorrect
`User` doesn't implement `getUserIdentifier()` as required, and `UserInterface` isn't strong enough to guarantee `getInstitution()` is available. We need to think about what we're trying to do here and fix the errors.

rob-3:
Based on #778; mark as ready when that is merged. Resolves #775


SimonRothUCF:
This looks good to me and I am not getting any errors. The only thing I noticed is that there is still one instance of `getUsername()` being called in `User.php`. If this is intended, it should be good to go.

SimonRothUCF:
Ignore that last comment, I misunderstood the commit message. Going to go ahead and merge.

PR #794, opened 2022-03-02
Adding Version Number to About Page

atarisafari:


assets/js/Components/AboutPage.js
```diff
@@ -131,6 +131,9 @@ class AboutPage extends React.Component {
             })}
           </ToggleDetails>
         </View>
+        <View as="div" textAlign="end">
+          <Text weight="light">{this.props.t('label.version')} {this.props.settings.versionNumber}</Text>
```
cidilabs, line 135: @atarisafari Do you know if UCF Open has `label.version` in its en.json?

PR #795, opened 2022-03-07
adding info for embedTagDetected and iframenothandled

atarisafari:


.env
```diff
@@ -64,7 +64,9 @@ PHPALLY_SUGGESTION_RULES="
     ObjectTagDetected,
     ParagraphNotUsedAsHeader,
     PreShouldNotBeUsedForTabularValues,
-    RedirectedLink
+    RedirectedLink,
+    EmbedTagDetected
```
rob-3, line 68: Should there not be a comma here?

PR #799, opened 2022-03-21
WIP: Rewrite and simplify docker configuration

rob-3:


rob-3:
I was unable to get this fully working on my AWS testbed, but I believe this is due to the fact I don't have a testing domain or https configured, which Canvas really doesn't like. I think it represents a step forward nonetheless

rob-3:
> Everything is pretty solid: I would recommend changing the . in your dockerfile to include the work directory, to keep consistency with the later parts of the build process. `COPY --from=node-builder /app/node_modules/ ./node_modules/ COPY --from=php-builder /app/vendor/ ./vendor/`(lines 28-29) is an example that uses the workdir compared to the earlier parts of the build which used a . to signify the working directory.

If you could reference the exact lines or even write a patch that shows exactly what you mean by this I would really appreciate it. I'm not the most familiar with Docker and not certain I fully understand the WORKDIR as it is.

rob-3:
This is blocked by the following issue:
```
Uncaught PHP Exception Error: "Call to a member function getInstitution() on null" at /var/www/html/src/Controller/AuthController.php line 123 {"exception":"[object] (Error(code: 0): Call to a member function getInstitution() on null at /var/www/html/src/Controller/AuthController.php:123)"} []
```
The root cause of this is authentication failure, because we are trying to call `user->getInstitution()` on a `null` user, which Symfony says is only provided when unauthenticated. Authentication appears to work fine on Heroku, so I'm unsure what is going wrong here. Related files are `src/Security/SessionAuthenticator.php` and `src/Service/SessionService.php`.

bagofarms:
We ended up going a different direction with the Docker setup.  Thank you for all your hard work on this PR.  It informed our new setup quite a bit!

PR #806, opened 2022-04-19
Bring changes from Cidilabs docker setup to UCF UDOIT

sgcooper78:
This is to bring changes made inside of Cidi Labs Docker setup to UCF.
There are still many questions. 

- If this is to be cloud agnostic then we really don't need to install AWS CLI tools.
- I am also not sure but for cloud versions you can do the builds in the Dockerfiles but Docker Compose and any files mounts get mounted after builds which will overwrite things like composer/yarn installs.
- The entrypoint script that is ran was changed to make sure you can run commands during startup of the container. Running it as a RUN defeats the purpose of the script and makes it part of the BUILD phase instead of in the actual container.

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -1,12 +1,12 @@
 ###> symfony/framework-bundle ###
-APP_ENV=prod
+APP_ENV=dev
 
 ###> doctrine/doctrine-bundle ###
 DATABASE_URL=mysql://root:root@db:3306/udoit3
 
 ###> base url ###
 # Base URL for client callbacks 
-BASE_URL="https://127.0.0.1:8000"
+BASE_URL="https://localhost"
 ###> base url ###
 
 APP_LMS=canvas
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,5 @@ yarn-error.log
 /phpunit.xml
 .phpunit.result.cache
 ###< phpunit/phpunit ###
+
+/database*
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,20 @@
-FROM php:7.4-fpm
+FROM php:8.1-fpm
 ARG ENVIRONMENT_TYPE
+ARG LIBRE_INSTALL
 
 #Install dependencies and php extensions
 RUN apt-get update && apt-get install -y \
-        git \
         libfreetype6-dev \
         libjpeg62-turbo-dev \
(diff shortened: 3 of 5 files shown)


rob-3:
From what I have learned, the idiomatic way to use the Dockerfile is to try to move as much of the build-time work into the Dockerfile as possible. We can and should do things like `yarn install` inside the Dockerfile to take advantage of caching build stages and to improve reproducible builds across machines. I have an alternate implementation cooking over in the `docker-wip` branch that is attempting to do some of these things and reduce the amount of code in custom shell scripts.

That being said, these improvements do look worthwhile, and I will try to take a look at them this week. Thanks for the PR!

sgcooper78:
@rob-3 You are correct, however this is the thing I noted with working with docker compose. At the end docker compose mounts your file system to /var/www/html and overwrites your changes. Making the special case for doing it just on local in the entrypoint/startup. Cidilabs moved to a Dockerfile for AWS and a Dockerfile.local for local because of this, you could also just have an if statement in the Dockerfile. We only run the composer install & yarn install on local now.

rob-3:
> At the end docker compose mounts your file system to /var/www/html and overwrites your changes. 

Yes; I ran into this also. What I have been doing is a more granular volume mount like this:

```docker
volumes:
  - ./src:/var/www/html/src
  - ./config:/var/www/html/config
  - ./public:/var/www/html/public
  - ./.env:/var/www/html/.env
  - ./.env.local:/var/www/html/.env.local
  - ./bin:/var/www/html/bin
  - ./templates:/var/www/html/templates
```

The main downside is that it is easy to forget files by mistake. I couldn't find a way to do a "mount everything *except* `node_modules/` and `vendor/`, but that feels like it would be the best way, if possible.

> Cidilabs moved to a Dockerfile for AWS and a Dockerfile.local for local because of this, you could also just have an if statement in the Dockerfile.

Is this `Dockerfile.local` public anywhere? I'd be interesting in trying to merge it here.

sgcooper78:
That's not a bad idea honestly. I don't think the Dockerfile.local is available anywhere however it's basically what is in this commit in the regular Dockerfile and the udoit-ng.sh with the else in the $ENVIRONMENT_TYPE doing the installs.

bagofarms:
We're going a different direction with the open source Docker setup, so I'm going to close this PR

PR #809, opened 2022-04-25
Add Typings Everywhere

### LINKED ISSUE #763: Add typings everywhere
PHP 7+ has solid support for in-code type hints for scalar (non-array) types. We should update function declarations with typings and remove docstring comments with types. For example:

```php
/**
   * API call GET
   *
   * @param string $url
   * @param array $options
   * @param integer $perPage
   * 
   * @return LmsResponse
   */
  public function apiGet($url, $options = [], $lmsResponse = null)
```
should become
```php
  public function apiGet(string $url, array $options = [], LmsResponse $lmsResponse = null): LmsResponse
```

If there is an existing docstring with something like `@param SomeType[]`, that cannot currently be expressed in PHP and is fine to leave.

Functions that don't return should be marked as `void`.

SimonRothUCF:
Closes #763

### DIFF
--- a/src/Controller/DashboardController.php
+++ b/src/Controller/DashboardController.php
@@ -90,13 +90,7 @@ class DashboardController extends AbstractController
         ]);
     }
 
-    /**
-     *
-     *
-     * @param Course $course
-     * @return void
-     */
-    protected function getSettings(Course $course)
+    protected function getSettings(Course $course): Array
     {
         /** @var User $user */
         $user = $this->getUser();
--- a/src/Controller/DashboardController.php
+++ b/src/Controller/DashboardController.php
@@ -90,7 +90,7 @@ class DashboardController extends AbstractController
         ]);
     }
 
-    protected function getSettings(Course $course): Array
+    protected function getSettings(Course $course): array
     {
         /** @var User $user */
         $user = $this->getUser();
--- a/src/Entity/ContentItem.php
+++ b/src/Entity/ContentItem.php
@@ -185,7 +185,7 @@ class ContentItem implements \JsonSerializable
         return $this;
     }
 
-    public function getIssues(): Collection
+    public function getIssues(): Collection|Array
     {
         return $this->issues;
     }
--- a/src/Entity/ContentItem.php
+++ b/src/Entity/ContentItem.php
@@ -86,7 +86,6 @@ class ContentItem implements \JsonSerializable
 
     // Public Methods
 
-    // Serializes ContentItem into JSON.
     public function jsonSerialize(): array
     {
         return [
--- a/src/Entity/ContentItem.php
+++ b/src/Entity/ContentItem.php
@@ -184,7 +184,7 @@ class ContentItem implements \JsonSerializable
         return $this;
     }
 
-    public function getIssues(): Collection|Array
+    public function getIssues(): Collection | array
     {
         return $this->issues;
     }
--- a/src/Entity/ContentItem.php
+++ b/src/Entity/ContentItem.php
@@ -86,10 +86,7 @@ class ContentItem implements \JsonSerializable
 
     // Public Methods
 
-    /**
-     * Serializes ContentItem into JSON.
-     * @return array
-     */
+    // Serializes ContentItem into JSON.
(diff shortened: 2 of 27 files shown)


src/Entity/Course.php
```diff
@@ -237,10 +231,7 @@ public function removeContentItem(ContentItem $contentItem): self
         return $this;
     }
 
-    /**
-     * @return Collection|Report[]
-     */
-    public function getReports(): Collection
+    public function getReports(): Collection|Array
```
rob-3, line 234: This and other instances of `Array` should probably be `array` I think. Not sure if it makes a difference, but good style. Also, put a space like `Collection | array`

rob-3 (changes requested):
Looks good, but I think a few style changes are in order. I also still need to test this and make sure nothing breaks.

src/Controller/DashboardController.php
```diff
@@ -90,13 +90,7 @@ public function index(
...
     }
 
-    /**
-     *
-     *
-     * @param Course $course
-     * @return void
-     */
-    protected function getSettings(Course $course)
+    protected function getSettings(Course $course): Array
```
rob-3, line 93: array

src/Entity/ContentItem.php
```diff
@@ -86,10 +86,7 @@ public function __construct()
 
     // Public Methods
 
-    /**
-     * Serializes ContentItem into JSON.
-     * @return array
-     */
+    // Serializes ContentItem into JSON.
```
rob-3, line 89: Honestly, you can drop this comment entirely. It doesn't add any information

src/Entity/ContentItem.php
```diff
@@ -188,10 +185,7 @@ public function setBody(?string $body): self
         return $this;
     }
 
-    /**
-     * @return Collection|Issue[]
-     */
-    public function getIssues(): Collection
+    public function getIssues(): Collection|Array
```
rob-3, line 188: `Collection | array`

src/Entity/Course.php
```diff
@@ -195,9 +192,6 @@ public function setLastUpdated(?\DateTimeInterface $last_updated): self
         return $this;
     }
 
-    /**
-     * @return ContentItem[]
-     */
     public function getContentItems(): Array
```
rob-3, line 195: `array`

src/Entity/Course.php
```diff
@@ -321,9 +312,6 @@ public function getUpdatedReport()
         return $report;
     }
 
-    /**
-     * @return FileItem[]
-     */
     public function getFileItems($activeOnly = true): Array
```
rob-3, line 315: `getFileItems(bool $activeOnly = true): array`

src/Entity/FileItem.php
```diff
@@ -235,10 +235,7 @@ public function update($file): self
         return $this;
     }
 
-    /**
-     * Serializes ContentItem into JSON.
-     * @return array
-     */
+    // Serializes ContentItem into JSON.
```
rob-3, line 238: Drop this comment also

src/Entity/User.php
```diff
@@ -290,10 +280,7 @@ private function decryptData($encrypted)
         return sodium_crypto_secretbox_open($encrypted_text, $nonce, $key);
     }
 
-    /**
-     * @return Collection|Report[]
-     */
-    public function getReports(): Collection
+    public function getReports(): Collection|Array
```
rob-3, line 283: `Collection | array`

src/Lms/Canvas/CanvasLms.php
```diff
@@ -164,15 +164,8 @@ public function updateCourseData(Course $course, User $user)
...
-     * Get content from Canvas and update content items
-     *
-     * @param Course $course
-     * @param User $user
-     *
-     * @return ContentItem[]
-     */
-    public function updateCourseContent(Course $course, User $user)
+    // Get content from Canvas and update content items
+    public function updateCourseContent(Course $course, User $user): Array
```
rob-3, line 168: `array`

rob-3 (approved):

Issue #788, opened 2022-02-23
Clean up JS


rob-3:
We should take a look through the JavaScript side of the project and clean up things like unused variables and fix any unjustified warnings.

We could also consider updating to function components, but that may be more of a long-term goal. The class components we have are still 100% supported by React.

rob-3:
This involves cleaning up some of the code in `assets/js/`. The main items are:

- [ ] Rewriting the classes in `assets/js/Services` into plain functions. There is no good reason for these classes with no fields or state to exist in JavaScript, which has free functions. This is basically turning things like:
```js
export default class Ufixit {
  returnIssueForm(activeIssue) {
    if (activeIssue) {
      if (UfixitForms.hasOwnProperty(activeIssue.scanRuleId)) {
        if (!activeIssue.sourceHtml.includes("data-type-module")) {
          return UfixitForms[activeIssue.scanRuleId]
        }
      }
    }

    return UfixitReviewOnly
  }
}
```
into
```js
export function returnIssueForm(activeIssue) {
  if (activeIssue) {
    if (UfixitForms.hasOwnProperty(activeIssue.scanRuleId)) {
      if (!activeIssue.sourceHtml.includes("data-type-module")) {
        return UfixitForms[activeIssue.scanRuleId]
      }
    }
  }

  return UfixitReviewOnly
}
```
and updating the usages and imports elsewhere. You will also get to remove a bunch of instances of `this` after this is done.
- [ ] Removing any unused imports
- [ ] Removing any unused variables/parameters
- [ ] Consider using [public fields syntax](https://dev.to/guin/learn-public-class-field-syntax-by-refactoring-a-react-component-njh) for class components to avoid the need for `.bind()` everywhere 
- [ ] (Possibly) Updating to function components. Someone with working knowledge of class-based and function-based React components will need to do this

PR #811, opened 2022-04-27
Remove unused JS imports

SimonRothUCF:
This is part of issue #788, it removes unused imports. It also removes instances of `eval()`, but this was already done in PR #810.


bagofarms (approved):
I had to drop the changes to contrast.js because the structure of that file is totally different now, but it's no longer using evals, so we're good.  Thanks!

PR #816, opened 2022-05-18
Creates docker containers and services for UDOIT via nginx

ssilverm:


ssilverm:
Instructions

```
git clone git@github.com:ucfopen/UDOIT.git udoit3                                                       
cd udoit3
git checkout issue/815-docker-nginx   
```

# .env.local
base_url = http NOT https

# .env
I changed the Jwks to point to *.test.*

`docker-compose -f docker-compose.nginx.yml up `
Wait a while…

docker-compose -f docker-compose.nginx.yml run php-fpm php bin/console doctrine:database:create

docker-compose -f docker-compose.nginx.yml run php-fpm php bin/console doctrine:migrations:migrate

Insert data into institution table.

bagofarms:
When running the doctrine:database:create line, I get the following error:
```
[critical] Uncaught Error: Class "Symfony\Bundle\WebProfilerBundle\WebProfilerBundle" not found
```

If I change `ENV` to `prod` in the `.env.local` file, I get the following error when running the doctrine:database:create line:
```
Symfony\Component\HttpKernel\EventListener\DebugHandlersListener::__construct(): Argument #6 ($scope) must be of type bool, Symfony\Component\HttpKernel\Debug\FileLinkForma
  tter given, called in /var/www/html/var/cache/prod/ContainerRWT9OZD/App_KernelProdContainer.php on line 2995
```

bagofarms:
Ok, it was using an old composer.lock file.  I also added gd and supporting packages to the PHP container so that it can use gd during runtime.

bagofarms:
I finally got it working!  The problem was some sneaky `https`es in the LTI Key.  We need to add instructions for anyone doing local development to dig through every part of the LTI Key and change all URLs for their local machine to `http`

PR #820, opened 2022-06-02
Update Canvas Pages scopes from :url to :url_or_id

escull75:
In IU's Canvas instance, we were getting API "invalid scope" errors when scopes were enforced on the API client. This change aligns the Pages scopes with what we have in our Canvas instance, and fixed the invalid scope issue.

### DIFF
--- a/INSTALL_CANVAS.md
+++ b/INSTALL_CANVAS.md
@@ -39,51 +39,52 @@ Provide values for the following fields:
 ### Scopes
 We strongly recommend you enforce scopes with your API key. The following scopes must be enabled for UDOIT to work.
 
-* Accounts
-    * url:GET|/api/v1/accounts
-    * url:GET|/api/v1/accounts/:id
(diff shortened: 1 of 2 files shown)


jonespm:
I'm seeing this same change.
image

Looks like it was recent,[ just from about a month ago. ](https://github.com/instructure/canvas-lms/commit/ceed52fc49b33377517dc1e53ec2f9b1bda2bc35)

Also the documentation on https://github.com/ucfopen/UDOIT/blob/dev/v3-3-0/INSTALL_CANVAS.md#scopes has these scopes

Additionally it has a scope not in this php file of 
`url:GET|/api/v1/groups/:group_id/pages/:url`

Which could probably be removed if it's not used

mrash2:
We started receiving "Authentication problem: Access Denied. invalid_scope" error upon launching UDOIT from Canvas LTI as of yesterday. It was working prior and I imagine the update is due to Instructure's changes to the scopes. What do you suggest we do to resolve this?

PR #835, opened 2022-07-28
Attempt to add PostgreSQL support in Docker. Currently broken

### LINKED ISSUE #834: Update Docker code to support PostgreSQL
The current commented-out code for using PostgreSQL instead of MySQL in Docker is not functional.

bagofarms:
This fixes #834

### DIFF
--- a/build/nginx/Dockerfile.php.pdo.pgsql
+++ b/build/nginx/Dockerfile.php.pdo.pgsql
@@ -1,0 +1,5 @@
+FROM php:8.1-fpm
+
+# PHP extensions
+RUN apt-get update && apt-get install -y libpq-dev libpng-dev zlib1g-dev git unzip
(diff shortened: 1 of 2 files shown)


jonespm:
Maybe have this as a separate docker-compose file if you were going to add this? 

Like have a 
`docker-compose.yml` (Whatever UCF feels the defaults should be, maybe Postgres+Nginx)
`Then docker-compose-nginx-mysql.yml `(For MySQL+Nginx for instance, if the default is Postgres)

Locally we're going to run UDOIT on MySQL so would be nice to have the option to run on the same DB as our production servers.

bagofarms:
Closing this because of how stale it is. I think @jonespm had a good idea, so maybe someone else can take a stab at it in the future.

PR #836, opened 2022-08-04
changed scopes to reflect canvas changes

atarisafari:
Canvas updated their scopes so we needed to make some changes in order to avoid failed API call issues.


bagofarms:
We have already made this change in 3.3.0, which I'm trying to get released this week.

PR #838, opened 2022-08-08
Add excel description and spanish translation

### LINKED ISSUE #832: Add description text for XLSX files
There is currently no translation if an XLSX file is reviewed.  It shows up like this
Screen Shot 2022-07-28 at 4 59 59 PM

But it should look something like this
Screen Shot 2022-07-28 at 5 00 11 PM

In order for this to be considered "done", there should be both an English and a Spanish translation.

bagofarms:
Closes #832


bagofarms:
Huge thanks to @emanuelGitCodes and @danielmolares for their help on this!

Issue #847, opened 2022-09-08
Create Helm chart deployment for UDOIT


bryancpowell:
[Helm](https://helm.sh/) is the standard way to deploy applications to a Kubernetes cluster in many institutions. A Helm chart for UDOIT would enable deployments to Kubernetes to be greatly simplified. Since the release of 3.3.0 includes a fully containerized build of UDOIT, this process is much more tenable.

The Helm chart should be part of this project, with chart publishing handled through GitHub Actions. This process is described in the document [Chart Releaser Action to Automate GitHub Page Charts](https://helm.sh/docs/howto/chart_releaser_action/).

This feature will be handled by Miami University/(miamioh.edu).

dmols:
Based on the comments in the #891 pull request, it doesn't seem like we'll need to make changes on our end. Closing this issue.

PR #848, opened 2022-09-09
Add Helm chart and deployment method

bryancpowell:
This pull request will close issue #847


.env
```diff
@@ -152,9 +151,7 @@ DATE_FORMAT="Y-m-d"
 
 ###> base url ###
 # Base URL for client callbacks
-BASE_URL=""
-# Webpack public path
-WEBPACK_PUBLIC_PATH="/build"
```
bagofarms, line 157: Does Helm require that you set these valuse in .env rather than .env.local? Removing this variable will default the build path to `/build` (as seen in https://github.com/ucfopen/UDOIT/blob/dev/v3-3-1/webpack.config.js#L13). If `BASE_URL` is set to `http://127.0.0.1:8000/udoit3`, then `WEBPACK_PUBLIC_PATH` needs to be set to `/udoit3/build`

Or do these changes even matter, since you're configuring everything through the `values.yml`?

INSTALL.md
```diff
@@ -16,6 +16,10 @@ The system requirements depend on how you install UDOIT.  If you use Docker, the
 * Node v16 is supported; other versions may work
 * Yarn
 
+### Helm Method (for deployment to Kubernetes)
+
+See [INSTALL-HELM.md](INSTALL_HELM.md) for details on prerequisites for this installation method.
```
bagofarms, line 21: Should it be this instead?
```suggestion
See [INSTALL_HELM.md](INSTALL_HELM.md) for details on prerequisites for this installation method.
```

INSTALL.md
```diff
@@ -119,6 +123,10 @@ To build the JavaScript files for production, run the command:
 
     yarn build
 
+## Helm Installation (for deployment to Kubernetes)
+
+See [INSTALL-HELM.md](INSTALL_HELM.md) for guidance on this method.
```
bagofarms, line 128: Would it make more sense to just redirect them to `INSTALL_HELM.md` in the beginning? If they've gotten to this point, have they already gone down the wrong path?

bagofarms (changes requested):
I just had a few issues and questions related to the .env and documentation.

INSTALL.md
bryancpowell: I am ok with either. I think because it was added last I put it at the bottom IIRC. It is a little weird because both this and the link above for the prerequisites just redirect to INSTALL_HELM.md. Let me know what you prefer and I will change it.

.env
bryancpowell: I must admit I do not recall why this was removed/changed at all. Looking. at the commits, it seems the `WEBPACK_PUBLIC_PATH` variable was not in the original branch we worked off of? https://github.com/ucfopen/UDOIT/pull/848/commits/9519c9ec4b512197c70402b84fe9842e9fd6d482

I may be misunderstanding.

In any case for our purposes, they do not actually matter. I will fix this to match what is there now

INSTALL.md
donkidd: I think this is updated now

INSTALL.md
bryancpowell: If you prefer, we can just put a single header at the top that says refer to `INSTALL_HELM.md` to install on Kubernetes. I was trying to match the organization of the other install methods but the way it is now is redundant as you noted.

bryancpowell: Let me know what you think in the last commit.

bagofarms:
I resolved the conflicts in a new branch, but I was unable to merge those changes into your fork. I'm closing this PR and making a new one.

PR #853, opened 2022-10-10
issue/644 - Issue navigation not always working

jonespm:
This seems to be working reliably for me in initial testing. I wonder if this was the initial idea for this @bagofarms? It looks like this has been in there for 2 years from a commit by @webchuckweb. I'm surprised you haven't seen it come up?

### DIFF
--- a/assets/js/Components/ContentPage.js
+++ b/assets/js/Components/ContentPage.js
@@ -79,7 +79,7 @@ class ContentPage extends React.Component {
 
   static getDerivedStateFromProps(props, state) {
     const stateActiveIssue = state.activeIssue
-    const propsActiveIssue = stateActiveIssue && props.report.issues[stateActiveIssue.id]
+    const propsActiveIssue = stateActiveIssue && props.report.issues.find(x => x.id === stateActiveIssue.id);
     if(propsActiveIssue && propsActiveIssue.status !== stateActiveIssue.status) {
       return {
         activeIssue: propsActiveIssue


bagofarms:
This change causes a different issue. I fixed issue 121, and after fixing, it became issue 116. I fixed the next issue, and it also became 116. That ID seems to be the first issue for that particular content item (a page in this case).

I switched back to the production code and reset the test course, and I get a different problem. When I fix issue 121, it becomes issue 1.

jonespm:
Thanks for testing. I'll try it out more tomorrow. I didn't test the edge
cases or fixing issues while navigating. Seemed to good/easy to be true.
But I'm at least able to replicate this more frequently now.

jonespm:
From what I can see the problem you describe _seems_ to be there with or without the fix on this PR. Resolving issues and then performing navigation introduces issues. 

In both cases when you click resolve it marks the issue as resolved but leave it in the navigation list. However when you click resolve on another issue it removes the previous resolved issue and sometimes causes the order to be messed up.

I feel like it should just leave the resolved issues in the navigation list until the modal is closed otherwise the user can't easily undo this action. 

Is this how it would be expected to work, or should the the resolved issue be removed from the filtered list immediately?

jonespm:
So here's the problem I'm seeing with this and I'm getting closer

filteredRows are getting updated every render cycle with

`const filteredRows = this.getFilteredContent();`

As issues are resolved, this FilteredContent is being updated, it's not static. This is messing up the list.

It looks like it's a problem with the code that checks the recentlyUpdated and recentlyResolved status and is removing it from the list. This filter *seems* problematic to me to use as it's navigating through the list when Active is the default filter checked and it doesn't actually update until the second action is taken. 

I feel like maybe this can be improved if *maybe* Active isn't checked by default and if it is the issue that's resolved is removed from the list immediately. I'm not sure why it isn't quite yet. 

```
      // Check if we are filtering by issue status
      if (!issue.recentlyUpdated && !issue.recentlyResolved) {
        if (filters.issueStatus.length !== 0 && !filters.issueStatus.includes(issueStatusKeys[issue.status])) {
          continue;
        }
      }
```

jonespm:
I _feel_ working on this that there are two separate issues and this fixes the one related to navigation not working correctly. 

The other one @bagofarms mentioned related to resolving/fixing issues **and** navigation is present in some form with or without this fix. We are also looking at that.

Issue #854, opened 2022-10-21
UDOIT Admin tool shows empty list for courses and reports


zqian:
The Admin tool shows empty list for Courses and Reports (see screenshot below)

![udoit-admin](https://user-images.githubusercontent.com/359944/197252532-3ea5f93b-ca2d-417d-b89c-0449f6aba7db.png)

The list inside Users tab works fine.

zqian:
Close the issue after the code merge.

PR #855, opened 2022-10-24
#854 flatten items from multiple array and merge into one single array

zqian:
Addressing #854 

The paged Canvas terms API call returns multiple arrays. 

This fix is to flatten the arrays and merge all items from all arrays. Multiple arrays are returned because UM has 101 academic terms returns, so that becomes two arrays since the default array size is 100 in Canvas api calls.


bagofarms (approved):
Tested and works fine in my instance. Thank you for the contribution!

PR #869, opened 2022-12-13
Updated phpally to 1.2.1

### LINKED ISSUE #785: Feature: add logic for superseding/conflicting text color attributes
Many of our course pages use text headings with white-on-gray text. For whatever reason (it was done before my time at the institution), this was done by giving the H3 tag a gray background and a span tag within it a white text color. 

Example:
`<h3 style="background: #58595B; text-align: left; padding-left: 5px;"><span style="color: #ffffff;"><strong>Basic Elements of the Course</strong></span></h3>`

This renders properly as white-on-gray in the browser. However, UDOIT picks up the color attribute of page and/or H3 text and flags this as a contrast error. Is it possible to add logic such that UDOIT recognizes that the color of the span tag supersedes the regular defined color for page text?

image

We can fix this manually by moving the color attribute from the span tag into the H3 (which is how I would have done it to begin with - but again, vast numbers of pages have already been created that way). It would be nice if UDOIT saw that the span color overrides the H3 color.

dmols:
Fixes #785 .

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "ext-ctype": "*",
         "ext-iconv": "*",
         "ext-sodium": "*",
-        "cidilabs/phpally": "~1.1.1",
+        "cidilabs/phpally": "~1.2.1",
         "composer/package-versions-deprecated": "1.11.99.3",
         "doctrine/annotations": "^1.0",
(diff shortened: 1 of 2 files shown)


bagofarms (approved):
Tested and working. Thanks!

PR #881, opened 2023-01-12
Modifications to support Docker Compose V2

### LINKED ISSUE #851: env_file in docker-compose not overriding. Remove/comment duplicates from .env
In the docker-compose.yml there are two files defined .env and .env.local.

However any changes made to .env.local aren't overriding the values in .env.

It looks like this is expected behavior.
https://forums.docker.com/t/combining-docker-compose-override-files-with-mutliple-env-files/119895/3
https://github.com/docker/compose/issues/7326#issuecomment-1072390032

I think it should either just be one file that we make the changes in, or the changes should be removed/commented out of the default .env file. Seems easier to just have a .env.example? 

To test just start this up locally with docker compose and make a change to the .env.local. It will still be using the value from .env, like BASE_URL. 

Also some values like JWK_BASE_URL are only defined in the .env and those may need to be changed or just removed entirely.

bagofarms:
Removes the tracked .env file, sets up a .env.example file, removes the .env.local.example file, and updates the Docker Compose and CI yml files to work with the new setup. Fixes #851


docker-compose.nginx.yml
```diff
@@ -26,42 +26,38 @@ services:
...
         - .env
-        - .env.local
   yarn:
       image: node:16
       volumes:
-          - ./:/app
+        - ./:/app
       env_file:
         - .env
         - .env.local
```
jonespm, line 63: This .env.local needs to be removed. 
```suggestion
```

jonespm:
Looks like this this is working for me

A few minor things I'd update in the documentation in addition to other change I suggested since this is related to Docker Compose.

* `docker compose` is the [built in command preference](https://docs.docker.com/get-started/08_using_compose/) now over the separate command `docker-compose`. There's a few places in the INSTALL.md I'd remove the hyphen. 
* `.env.local` is mentioned once still in INSTALL.md and should probably change to `.env`
* `.env.local` is mentioned twice in INSTALL_CANVAS.md and should probably change to` .env`

Thanks!

jonespm (approved):

Issue #887, opened 2023-02-13
Recently updated state does not persist past the latest resolved or fixed issue?


ssciolla:
When fixing and resolving issues in the `UFixitModal` when filtered to "Active Issues" (the default), if the user fixes or resolves multiple issues, only the latest updated issue stays in the modal. See the video below for an example:

https://user-images.githubusercontent.com/35741256/218348701-4fc74887-01b5-41d0-a0c6-cf87e1f26400.mov

I believe this is because the `report` from the backend supersedes the frontend `report` in this handler method, and the extra logic that follows to keep the last resolved issue in the modal only affects the latest issue.

https://github.com/ucfopen/UDOIT/blob/d15538d2826e84173b6bbf7da975d2aae1a90a0f/assets/js/Components/App.js#L219-L228

For people familiar with the design decisions with version 3, is this deliberate? There are small changes we could make here to keep all fixed or resolved issues in the modal until it has been closed. I think the current behavior is somewhat confusing and it would make sense to keep anything changed while the modal has been open in view, so users can review (or un-do) their work.

PR #888, opened 2023-02-14
Remove `ContentPage.getDerviedStateFromProps` to fix navigation bug (#644)

### LINKED ISSUE #644: Issue navigation not working
Selecting an issue on the UFIXIT page, as well as the "Next Issue" and "Previous Issue" buttons, are all not working properly.  Instead of the proper issue, an issue that is marked as fixed will show up in its place, even if I have fixed issues filtered out. 

I believe that this is due to the links not being properly updated when an issue is marked as fixed.

This is happening on UDOIT 3.0.
Examples:
When I click on this issue, which is third on the list of issue:
image
This other issue pops up
image

When I try to navigate to the third issue through the Next or Previous Buttons
image
I reach the same issue
image

It is effectively impossible for me to reach that issue.

ssciolla:
This PR aims to resolve #644. It is currently a draft but I hope to open it for review shortly.

# Explanation

The `getDerivedStateFromProps` method in the `ContentPage` component was introduced in this commit: https://github.com/ucfopen/UDOIT/commit/baf5f0e301d41cf89cb51c87023bddf07bb668a9 My colleague from University of Michigan @jonespm did a good job analyzing and explaining the issue, but I'll do so here for completeness.

Inside this method, there is a usage of `issue.id` as an index for finding an issue in `report.issues` with the same ID. Order in this array and ID are not equivalent. The result is that, in scenarios where the ID is small enough that it could also be an index (if it's not small enough, in JavaScript, accessing an invalid index results in `undefined`), the method ends up comparing two different issues. If one of the issues is fixed or resolved (i.e. it has a different `status`), the issue from the incorrect indexing supersedes, leading to a jumping of indexes in the UI to a completely unrelated issue.

@jonespm proposed initially in PR #853 that we use `find` as a replacement for the logic here. That would likely work fine, but I haven't yet figured out what value this method adds. It's supposed to update the `activeIssue` state of `ContentPage` if what it receives from `props` is different from what it has in internal state. Okay, cool, but shouldn't that state already be updated by `ContentPage.handleActiveissue`, which is called in several places in `UFixitModal` and its children. Since this method hasn't been functioning as intended, I'm inclined to think it's not necessary at all, but let me know if I'm missing something here.

*Note*: there are other issues with the navigation that leads to index jumping, but usually it's just the result of recently state not being kept beyond the first (see #887) or within a content item grouping (something I think @taheralfayad was working and I will try to document). I feel these should be addressed separately.

# Testing instructions

I will add some testing instructions here shortly. You may find it useful to apply the following diff so that the issue ID is visible in the modal, in the header.

```
diff --git a/assets/js/Components/UfixitModal.js b/assets/js/Components/UfixitModal.js
index cfa1e03f..f05c43d4 100644
--- a/assets/js/Components/UfixitModal.js
+++ b/assets/js/Components/UfixitModal.js
@@ -106,7 +106,7 @@ class UfixitModal extends React.Component {
           <Modal.Header padding="0 medium">
             <Flex>
               <Flex.Item shouldGrow shouldShrink>
-                <Heading>{this.props.t(`rule.label.${activeIssue.scanRuleId}`)}</Heading>
+                <Heading>{this.props.t(`rule.label.${activeIssue.scanRuleId}`)} (issue ID: {activeIssue.id})</Heading>
               </Flex.Item>
               <Flex.Item>
                 <CloseButton
```

# Resources
- https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

### DIFF
--- a/assets/js/Components/ContentPage.js
+++ b/assets/js/Components/ContentPage.js
@@ -77,17 +77,6 @@ class ContentPage extends React.Component {
     }
   }
 
-  static getDerivedStateFromProps(props, state) {
-    const stateActiveIssue = state.activeIssue
-    const propsActiveIssue = stateActiveIssue && props.report.issues[stateActiveIssue.id]
-    if(propsActiveIssue && propsActiveIssue.status !== stateActiveIssue.status) {
-      return {
-        activeIssue: propsActiveIssue
-      }
-    }
-    return null
-  }
-
   handleSearchTerm = (e, val) => {
     this.setState({searchTerm: val, filteredIssues: [], tableSettings: Object.assign({}, this.state.tableSettings, {pageNum: 0})});
   }


cidilabs:
Great job in identifying this issue. As you mention, this isn't THE answer to the modal index issue, but it's a good start.

cidilabs:
My only concern here is we don't know what we don't know. There was a use case for `getDerivedStateFromProps()` to be added in the first place. It was attempting to solve a problem, but we don't know what that problem was. I'll ask around to see if I can find out why it was added, but I agree that the logic is broken and should at least be fixed, if not removed.

ssciolla:
Thanks, @cidilabs , for taking a look and following up with your team on the purpose of this method. I appreciate it. I'm open to changing it to the `report.issues.find(...)` approach if we hear something that suggests it's needed.

I don't know this application as well as you all do, but my reading and experience with React applications suggest that the side effect nature of this method can make it somewhat hard to reason about versus relying on `ContentPage.handleActiveState`, which can be explicitly called in API callbacks. Even better, following the logic of the blog post I listed in the description, it may make sense to lift the `activeIssue` state up to the same level as the `report` in `App`, so that there's a "single source of truth" for this data and `ContentPage` becomes controlled with respect to `activeIssue`. I understand the latter suggestion would require a more significant refactor, so I've opted for the lighter weight solution here.

ssciolla:
I was taking a look at another issue, and I experimented a little with the alternative solution (basically what @jonespm had in #853).

```
  static getDerivedStateFromProps(props, state) {
    const stateActiveIssue = state.activeIssue
    const propsActiveIssue = stateActiveIssue && props.report.issues.find(i => i.id === stateActiveIssue.id)
    if (propsActiveIssue && propsActiveIssue.status !== stateActiveIssue.status) {
      return {
        activeIssue: propsActiveIssue
      }
    }
    return null
  }
```
 It may fix the issue jumping but I saw an issue with this approach; the loading indicator was not showing up while issues are being resolved. This seems to be because the issue gets marked with a resolved status right away, leading to a difference in state with what is being passed to props. Thus the props version wins out, and we also load the `pending` state.

https://user-images.githubusercontent.com/35741256/220978669-c86b4fcc-e720-42b7-ba6f-73b1d0d52034.mov

jonespm:
Thanks for looking into that @ssciolla , now I do remember that side effect and I didn't figure out why it was happening. The logic seemed more correct but when it was doing this status check I couldn't get it to sync up. 

I feel like I kept trying to modify the if statement in here as well to not switch to the props issue but couldn't see a combination that worked, so it just seemed like this needed something else.

dmols (approved):
Both the loading indicator error mentioned by @ssciolla and the unreachable issue on UFIXIT error (mentioned in #644) do not occur on my end. 

https://github.com/ucfopen/UDOIT/assets/70301465/d5921d7e-3919-4838-a15c-2546c6da0d85

All good to push!

ssciolla:
I don't have commit permissions (which is fine), please feel free to commit if you think it's good to go.

PR #891, opened 2023-02-23
Add Helm Chart and Deployment Method

### LINKED ISSUE #847: Create Helm chart deployment for UDOIT

bagofarms:
Fixes #847 
I had to create this because I was unable to push changes back to MiamiOH's repository.


bagofarms:
It looks like GitHub's PHP 8.1 runner updated to Node 17, which removed support for one of the crypto libraries we're using: https://github.com/webpack/webpack/issues/14532

bagofarms:
@bryancpowell We're still working on the github actions issue, but I wanted you to take another look at this and see if anything else needs to be updated from your branch. I can give you whatever access you need to update it. Also, keep in mind that this may need to be updated whenever we do new releases, and I will direct any questions about this to you since UCF doesn't use Helm.  

One more thing: I noticed that there's some `lorem ipsum` in `INSTALL_HELM.md`. Is that intentional, or can it be updated?

bagofarms:
Closing because MU has decided to maintain the helm chart in their own fork.

Issue #867, opened 2022-12-01
Re-resolving issues leads to spinning UI state, 404 in backend


ssciolla:
**Expected behavior**

Users are able to resolve and un-resolve issues repeatedly without issue. Furthermore, users will see an indication of errors in the UI.

**Actual behavior**

When users click try to resolve an issue that was previously unresolved, they see a loading indicator that goes on perpetually. An inspection of the Web console shows two errors:

```
POST https://ssci-udoit.ngrok.io/udoit3/api/issues/4139/resolve 404
Uncaught (in promise) SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
```

**Steps to reproduce**
1) Once UDOIT is loaded in a course, open an issue in the modal.
2) Check the resolve checkbox, wait for the action to occur.
3) Uncheck the resolve checkbox, wait for the action to occur.
4) Check the resolve checkbox again; observe that the action never completes.
**

**Commentary**

This problem appears to be because un-resolving an issue leads to a new issue being created (with a different ID) rather than an update to the existing issue. There might be a way to resolve this through significant refactoring in the backend, but the frontend could also be modified so that the former issue is replaced with the new issue, or at least an unresolved one. Regarding the lack of an error message, I'm not sure if there's an existing mechanism for reporting an HTTP error, or if there was some design choice related to this. The React application could probably benefit from a consistent error handling strategy if not.

PR #892, opened 2023-02-23
Keep recently changed issues in `UFixitModal` (#887)

### LINKED ISSUE #887: Recently updated state does not persist past the latest resolved or fixed issue?

ssciolla:
This PR aims to resolve #887. It is still a draft.

The changes here make it so multiple issues that have been fixed or resolved remain in the open `UFixitModal`. (In constrast, the code in `main` only keeps the most recently changed issue in the modal.) This is intended to assist users in reviewing their work (or potentially un-doing resolves, though that is currently problematic, see #867). It also prevents indexes in the bottom-left of the modal from shrinking by one (because previously fixed or resolved issues aren't being filtered out).

The recent state of the issues is set to `false` when the modal is closed; thus these "recently updated" issues will not appear in the modal if it is closed and re-opened. That's a departure from current behavior where the most recently updated issue would still appear if you close and reopen the modal with the same filters. My feeling is that clearing this on close would be the expected behavior, but that may be incorrect. Certainly it didn't make sense to keep all updated issues in recent state forever (or at least until the application is refreshed/re-scanned).

Some of these decisions/assumptions could benefit from being tested by UX folks.

### DIFF
--- a/assets/js/Components/App.js
+++ b/assets/js/Components/App.js
@@ -219,13 +219,13 @@ class App extends React.Component {
   handleIssueSave(newIssue, newReport) {
     const oldReport = this.state.report;
 
-    const report = {...oldReport, ...newReport};
+    const report = { ...oldReport, ...newReport };
 
     if (report && Array.isArray(report.issues)) {
       // Combine backend issues with frontend issue state
       report.issues = report.issues.map((issue) => {
         if (issue.id === newIssue.id) return newIssue;
-        const oldIssue = oldReport.issues.find(oldReportIssue => oldReportIssue.id === issue.id);
+        const oldIssue = oldReport.issues.find((oldReportIssue) => oldReportIssue.id === issue.id);
         return oldIssue !== undefined ? { ...oldIssue, ...issue } : issue;
       });
     }
--- a/assets/js/Components/App.js
+++ b/assets/js/Components/App.js
@@ -217,14 +217,20 @@ class App extends React.Component {
   }
 
   handleIssueSave(newIssue, newReport) {
-    let { report } = this.state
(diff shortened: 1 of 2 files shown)


ssciolla:
If this is going to be considered, I'd consider reviewing #888 first. Making that change may reduce other error noise, especially for people with new local builds or setups.

dmols:
I see this being especially helpful for bigger courses; aiming to manually resolve several issues with the filtering involved tends to slow down the process a good bit.

dmols (approved):
Changes look good. The `UFixitModal` works as intended and the issues can be manually resolved one after the other, without losing track of them. As mentioned, the resolved issues only go away upon closing the 'UFixitModal'. Clear to merge.

Issue #871, opened 2022-12-16
Adding a YouTube API key after scanning does not rescan videos


bagofarms:
If you set up UDOIT with an empty YouTube API key and scan a course, adding a YouTube API key and rescanning the course will not result in any videos without captions being detected. This is because those pages do not have new content in them, so UDOIT assumes it does not need to rescan them. Editing a page and saving it will trigger UDOIT to scan the videos using the new API key.

## Possible Solutions
1. Detect if the YouTube or Vimeo keys have changed, and force a full rescan of each course the next time UDOIT is visited inside that course.
2. Add an option to the Reports tab of the Admin panel that forces a full rescan of each course the next time UDOIT is visited inside that course.
3. Add a "Force Full Rescan" option. Maybe when someone clicks the "Rescan Course" option and no new content is found, a link to perform a full rescan could appear in the notification box.

PR #898, opened 2023-03-24
Add a full rescan option to UDOIT

taheralfayad:
With the intent to tackle #871 , I created a full course rescan option in UDOIT that deletes all the reports belonging to a course, and generates a new report in their stead. This solves the bug that we were encountering wherein adding a Youtube API key to the .env file after a course has been scanned does not automatically trigger a rescan of the course, and therefore, UDOIT misses issues that pertain to youtube captioning.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,13 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.operating-system }}
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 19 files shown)


dmols (approved):
Full rescan works as intended; fills issue list with new issues relating to captioning if a youtube API key is placed in the env file. Great addition! Clear to push.

dmols:
Changes look good! Added the spanish translation and now it should be good to go. I'll merge it after clearing the other PR.

dmols (approved):
Extra changes needed were added. Clear to push!

PR #900, opened 2023-03-31
Allow UDOIT to accept modern filetypes

### LINKED ISSUE #899: UDOIT does not accept modern filetypes for classic counterparts [docx, pptx, xlsx]
Upon using the review files tab in UDOIT, I found out that a file that has the .doc filetype can only accept .doc filetypes. Moreover, even files that are of type .docx can only accept .doc files. The same goes for pptx and xlsx files. This represents an issue for most people who nowadays use the .docx, .pptx, and .xlsx extensions. In addition to this bug, UDOIT returns a message that is vague and is not translated in the en.json file.
Screenshot 2023-03-27 at 1 25 20 PM

taheralfayad:
Fixes #899


dmols (commented):
Fix removes the error that prevented the "replacement" of files with one that has a modern extension, and vice-versa. However, it does not actually replace the file in question; it just adds the new file to the list of files for the course. This may be because there are references to said file that must be altered before the file can be fully replaced. Further discussion might be needed to decide how to tackle this problem.

dmols:
It may have happened because of clearing my cache, but doing the process again no longer adds the file to the list of files NOR replace the file in question. This is also an issue in `main` . As of now, UDOIT allows the user to replace a file only if it has the same name. If it has a different name, nothing happens but we get the 'File Replaced' confirmation dialog still.

taheralfayad:
I just verified, when a file is uploaded through the review files modal, it replaces the file that it is intending to replace. This pull request should be ready for merge.

dmols (changes requested):
Changes look good, great work! The only thing I would argue to add is a note mentioning that taking a look at the file and saving it as a docx, pptx, xlsx, etc, and re-uploading it, could fix the issue. Why I would add this note is because there's no way to replace a modern file with one that has the old extension. The alternative to this, is allowing them to replace modern filetypes with older filetypes, but we don't want to encourage that. 

Lastly, good call adding the message in the english translation. I can add a similar one to the spanish translation, if it needs it.

dmols:
Added spanish translation and made an issue (#940) to reword the modal message for finding incorrect/old filetypes. Clear to merge.

dmols (approved):

Issue #821, opened 2022-06-02
Invalid scope error on Canvas Pages scopes 


escull75:
In IU's Canvas instance, we were hitting invalid scope errors when enforcing scopes on the API client according to https://github.com/ucfopen/UDOIT/blob/main/INSTALL_CANVAS.md. I noticed that the scopes for Pages use :url_or_id, instead of :url, for their parameter. 

![image](https://user-images.githubusercontent.com/6439114/171682329-afccf6e2-5a00-4cad-aba0-5128daa6e310.png)

Changing these scopes in the CanvasLms.php file seemed to fix the issue in our instance. 

#820

jonespm:
Additionally we noticed that when trying to edit the API key that had these scopes defined, we were unable and it kept giving a error when clicking Save. I'm not sure if this is related to the changed names or not but does seem suspicious. I can't edit this key to change the scopes. No other API keys seem to have this problem.

I had to create a new API key, update the code to what's on the PR #820 and then change the API key in the UDOIT institution table to get this all to work again.

I guess if you didn't have scope enforcement on this probably wouldn't have been a problem. 

<img width="513" alt="image" src="https://user-images.githubusercontent.com/27447/174669705-3485bf9c-a611-4c3b-9b32-9db7da6291e8.png">

bagofarms:
Merged into dev/3.3.0 in #820

Issue #550, opened 2020-10-05
DB tables need to handle larger Canvas user IDs for Canvas trusted accounts


cidilabs:
When Canvas instances are set up as trusted accounts, the Canvas user ID changes from `1234` to `321000000000001234`. Currently the DB `reports` and `users` tables only support 10 digits for the Canvas user ID.

bagofarms:
Merged in #552

PR #902, opened 2023-05-05
Updates for UDOIT classic to fix a few issues

jonespm:
These were all _mostly_ for optional features but we happened to be using all of them.

I think the only required features for [Heroku 20](https://devcenter.heroku.com/articles/heroku-20-stack) is the east-desire change as it won't build on Heroku 20 without it and some of the PHP upgrades which were minor. This allows upgrading to the Heroku 20 stack since 18 is deprecated. We've been running this branch without any problems for a few months on Heroku 20.

Summary: AutoIdle (Heroku) support, update (optional) scoped tokens, update for trusted accounts when using a worker process. (Both of these aren't always standard). Update some dependencies for Heroku 20. This also removes support for PHP 7.3 and Heroku 20 is running 8.0.28. 

Heroku 20 is supported though April 2025. We intend to _finally_ be upgraded to UDOIT 3 by then. 😀

Note: I have not tested this on MySQL but the script looks correct. 

* Adds (optional) support for [AutoIdle](https://autoidle.com/): AutoIdle is an optional addon that will idle dev apps when they're not being used. It has saved us some money for some of our Heroku instances since Heroku doesn't idle the paid apps anymore. We've found that waking the app from AutoIdle sometimes requires a refresh so unfortunately not currently recommended for prod instances. This is something Heroku may add in the future too.
* Provides a fix for scopes on #821 if using scoped tokens. This was included only in UDOIT 3.  Also adds a message to check "Allow Include Parameters"
* Additional conversion for #550 since that was missing a column in job queue

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ UDOIT can be installed on your own existing servers, but we've also configured a
 
 To start the Heroku deployment process, you can click the button below, please note, that although this button eliminates much of the installation complexity, there are still some configuration steps that need to be followed, those steps are outlined in the [HEROKU.md Readme](HEROKU.md).
 
-<a href="https://heroku.com/deploy?template=https://github.com/ucfopen/UDOIT/tree/classic" title="Deploy to Heroku"><img src="https://www.herokucdn.com/deploy/button.svg" alt="Deploy to Heroku" title="Deploy to Heroku Button"></a>
+<a href="https://heroku.com/deploy?template=https://github.com/tl-its-umich-edu/UDOIT/tree/classic" title="Deploy to Heroku"><img src="https://www.herokucdn.com/deploy/button.svg" alt="Deploy to Heroku" title="Deploy to Heroku Button"></a>
 
 ## System Requirements
 * Apache or Nginx webserver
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -102,7 +102,7 @@ UDOIT can be installed on your own existing servers, but we've also configured a
 
 To start the Heroku deployment process, you can click the button below, please note, that although this button eliminates much of the installation complexity, there are still some configuration steps that need to be followed, those steps are outlined in the [HEROKU.md Readme](HEROKU.md).
 
-<a href="https://heroku.com/deploy?template=https://github.com/ucfopen/UDOIT/tree/classic" title="Deploy to Heroku"><img src="https://www.herokucdn.com/deploy/button.svg" alt="Deploy to Heroku" title="Deploy to Heroku Button"></a>
+<a href="https://heroku.com/deploy?template=https://github.com/tl-its-umich-edu/UDOIT/tree/classic" title="Deploy to Heroku"><img src="https://www.herokucdn.com/deploy/button.svg" alt="Deploy to Heroku" title="Deploy to Heroku Button"></a>
```
jonespm, line 105: Switch back Deploy to Heroku Button
```suggestion
<a href="https://heroku.com/deploy?template=https://github.com/ucfopen/UDOIT/tree/classic" title="Deploy to Heroku"><img src="https://www.herokucdn.com/deploy/button.svg" alt="Deploy to Heroku" title="Deploy to Heroku Button"></a>
```

bagofarms (approved):
I don't have an environment for testing this, but I want to merge this in asap.

PR #903, opened 2023-05-10
Remove "Fixed HTML was not replaced in LMS." message

### LINKED ISSUE #896: UDOIT Does Not Remove Suggestion/Error After Editing the Webcourse Page
While completing tickets, I ran into an issue regarding errors/suggestions that cannot be edited on UDOIT. Whenever I edited a page's content, for instance, changing a styled paragraph to a header or removing white highlight behind a text element, UDOIT would not recognize the page being altered upon reloading the LTI. This means that the error/suggestion would not be removed, and in the error/suggestion, the problematic HTML snippet before my change to it would still show up.

While I can occasionally get UDOIT to recognize the fix upon re-saving the page, sometimes UDOIT will just outright not recognize that the page was changed, making the error seem bugged. Trying to manually resolve the error will just give a red error alert.

Screenshot 2023-03-14 at 12 43 01 PM

Below is are images that show the warning still persisting after having changed the "Action Item" text from a paragraph to an h2 header.

Screenshot 2023-03-14 at 11 17 35 AM

Screenshot 2023-03-14 at 11 18 20 AM

Running Firefox if that makes a difference.

taheralfayad:
Fixes #896 

### **Reasoning behind removal:**

Currently, whenever UDOIT reports an issue and the user decides to fix the issue manually and changes the HTML causing that particular issue, UDOIT just returns a "Fixed HTML was not replaced in LMS" message, which would then prevent the issue from being resolved, due to this block of code in IssuesController.php:
```
// Add messages to response
$unreadMessages = $util->getUnreadMessages();
if (empty($unreadMessages)) {
```
This is causing issues because it prevents the workflow of fixing the issue and then marking it as resolved. Currently, mark as resolved just means ignore the issue, which is somewhat misleading to a person who wants to actually manually resolve an issue. The current workaround which we've developed is to first mark the issue as resolved and then change its HTML, but this change also allows us to switch the order of operations

### DIFF
--- a/src/Services/LmsPostService.php
+++ b/src/Services/LmsPostService.php
@@ -51,11 +51,6 @@ class LmsPostService {
 
         $replaceSuccess = $this->replaceContent($issue, $contentItem);
         if (!$replaceSuccess) {
-            $this->util->createMessage(
-                'Fixed HTML was not replaced in LMS. Please contact an administrator.',
-                'error',
-                $contentItem->getCourse()
-            );
             return;
         }
 


dmols (approved):
Works as intended. User can now manually resolve an issue without the error popping up, and the issue will not pop up again as if it were unresolved. Good work! Seems good to push.

dmols:
Merged `main` into branch just in case we saw conflicts. None were seen so we're good to merge in!

PR #910, opened 2023-07-07
WIP: Issue/718 create worker for scanning

### LINKED ISSUE #718: Create worker for scanning
Heroku times out if a process takes longer than 30 seconds (#713 ).  The official solution to this problem is to create a Worker that handles the long process.  In UDOIT 2.x, we created our own simple message queueing system.  UDOIT would place scanning requests in this queue, then the worker would look for the oldest incomplete request.

Rather than building our own system this time, we should look into using an existing mq system, like [RabbitMQ](https://medium.com/q-software/symfony-and-rabbitmq-86b14dd604b1) or [Symfony Messenger](https://symfony.com/doc/current/messenger.html#creating-a-message-handler)

### LINKED ISSUE #859: Display a updated progress of what's being scanned so user knows something is happening
In a course with a lot of content UDOIT 3 just displays a static message with a spinner like
image

If the scan takes awhile, longer than a few minutes because of a lot of content, there's no indication to the user anything is happening and they may be tempted to refresh. This may need some type of communication between the backend process that's running the process and the frontend. 

The previous version of UDOIT actually showed a live updated progress of what was being scanned which was nice.

image

This is somewhat related to #718 and #856

taheralfayad:
Fixes #859 
Fixes #718

### Motive:
The reasoning behind the changes made in this pull request was as follows: 

- UDOIT currently does not show progress updates on the scanning process, and the scanning process sometimes takes upwards of 40 seconds, meaning that users could interpret the behavior as unresponsive and reload the page if the course is too large.

- Heroku times out a process after 30 seconds of loading, which makes it impossible to scan large courses. As such, the refactoring of the scanning process as a worker process will allow for the scanning of larger courses.

### How the scanning process worked:

- Prior to the changes made in this pull request, the way that a course was scanned was that UDOIT would make a call to the controller at `/api/sync/{course}` and the controller then would call the `refreshLmsContent` method at `LmsFetchService.php`, which would then commence the scanning process. Finally, the controller would send back an updated version of the report.

### What changed:

- Instead of running the scanning process synchronously to the rest of the backend logic. The scanning process is now isolated in a Symfony Message Handler (more commonly known as an asynchronous background worker), and this was done using Symfony Messenger.


assets/js/Components/App.js
```diff
@@ -154,14 +189,88 @@ class App extends React.Component {
...
+        .then(() => this.pollBackgroundWorker())
+        .then(() => this.getReport())
     }
     this.forceUpdate()
   }
 
+  // As the worker process updates UDOIT with the latest content items, this method checks on whether or not
+  // This process has been completed
+  fetchUpdateData() {
+    fetch(`http://${window.location.hostname}:8000/udoit3/api/progress`)
```
jonespm, line 209: These can/should minimally be relative URL's. For example

`    fetch(`/udoit3/api/progress`)`

Otherwise they won't work on a hosted environment with SSL on regular ports. 

These might need to use the `BASE_URL` from the .env if this was changed from the default.

bagofarms:
Look into using Supervisor to see if it would be more reliable than `sleep` https://symfony.com/doc/current/messenger.html#supervisor-configuration

ewainberg:
Old PR, never really implemented.

PR #919, opened 2023-07-31
Add guardrails against transparent background color

### LINKED ISSUE #908: Transparent background-color styling throwing "Insufficient Text Color Contrast" error
The following code produces an "Insufficient Text Color Contrast With the Background" error even if there is no contrast issue:
`<p style="background-color: transparent;">`
Removing the `background-color` styling gets rid of the error but it's impossible to see it in the Rich Content Editor. UDOIT cannot change the color using the arrows to adjust value up and down in the UFIXIT editor (nothing happens if you use it), but it can successfully change the background color using the color picker drop-down.

taheralfayad:
Fixes #908 
Note: These changes solve this issue independent of the corresponding changes made to phpally in this [pull request.](https://github.com/ucfopen/phpally/pull/1)

### Issue:
As of recent, UCF instructional designers have been complaining about a bug that flags HTML elements with the `background-color: transparent` style as having insufficient contrast with their background. A [pull request](https://github.com/ucfopen/phpally/pull/1) was recently made to address this particular issue in phpally, making any elements with the transparent style inherit the background color of their parent element. The changes in the phpally pull request successfully got rid of the false positives, however, they could not address the scenario where there might actually be some color contrast issues. This is where this pull request comes in.

### Solution:
To address this issue, the changes made in this pull request will at first scan the HTML of any canvas page to look for any elements with the `background-color: transparent` style. If such elements are found, then UDOIT will change the HTML elements to inherit the background color of their parent element, which allows the user to change the true background color of the elements, and resolve any contrast issues resulting from them.

### DIFF
--- a/src/Lms/Canvas/CanvasLms.php
+++ b/src/Lms/Canvas/CanvasLms.php
@@ -2,6 +2,7 @@
 
 namespace App\Lms\Canvas;
 
+use DOMDocument;
 use App\Entity\ContentItem;
 use App\Entity\Course;
 use App\Entity\FileItem;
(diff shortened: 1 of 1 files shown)


dmols (approved):
Works as intended: the HTML is appropriately updated to not have a transparent background anymore and the issue no longer pops up in the UDOIT dashboard, since the HTML is updated as the course scan occurs. Great work! Good to push.

dmols:
We opted to find a better solution for this that doesn't involve altering HTML without the instructors' explicit approval. Will be making an issue to address this, for us to revisit in the future.

PR #938, opened 2024-04-17
Update Avoid Styles for Heading Text Issue Description

### LINKED ISSUE #923: High False Positive Rate with Avoid Using Styles for Document Structure
For version 3.3.2

Description: UDOIT seems to falsely report the Avoid Using Styles for Document Structure on text that is bolded but not a heading. 

Suggested/Potential Idea To Fix: Have a check to see if the word count is too long maybe greater than 16 characters (as that would not be any reasonable heading) and if so does not mark the text as a header that should not be bolded. I do not know the inner workings of UDOIT as much so this may not be applicable but is just a thought to consider.

Step by Step to Reproduce:
- Perform a UDOIT scan with content in on a page that has a header and then some bolded text and regular text.
- UDOIT will detect the bolded text and give an Avoid Using Styles for Document Structure despite it not being a header.

Noticed that it is semi-iffy on this after further testing for this bug report. Some instances it won't detect, other times it well. Sometimes if the text is inside a span, it gets flagged and follows same issue here. Other times like in the Example HTML provided it does not. Recommended when fixing an enhance dive into different test cases to see what is trigger it as the one example provide may not cover all false reports that this is affecting. 

Example HTML: 
```
<h2>This is A Header</h2>
<p><strong><span>This is some text that I have bolded because its really important! But unfortunately not as important as a header.</span><br /><br /></strong>text text text text text text text text</p>
<p>&nbsp;</p>
<h3>Header 236</h3>
<p><strong>test text more text more text more text more text more text more text more text</strong></p>
```
If you create a new page, copy and paste the code into the HTML editor, then save, publish, and run in UDOIT it should be flagged or at least was on my end. Again, does not cover all possibilities that UDOIT incorrectly flags bolded text.

The "test text more text more text more text more text more text more text more text" should be falsely reported by UDOIT with the Avoid Using Styles For Document Structure using the above HTML code in a page and as seen in the screenshot below.

Screenshot 2023-10-12 at 16 23 52

Severity: Low/Medium (can be worked around by marking as resolved, but still should not be the desired output of UDOIT)
Scale: Large, affects all users with html that has bolded text that is not a heading.

taheralfayad:
Kind of fixes #923. Change up the issue description to tell the user that UDOIT is not entirely sure that this is a usage of style for heading, but it is a good suspect.

An enhancing solution is absolutely needed for this issue.


translations/en.json
```diff
@@ -356,7 +356,7 @@
...
-  "rule.desc.ParagraphNotUsedAsHeader": "<p>Bold and Italics are used to emphasize text, whereas headings are used to define the structure of the document. Headings like <code>h1-h6</code> are extremely useful for non-sighted users to navigate the structure of the page, and formatting a paragraph to just be big or bold, while it might visually look like a heading, does not make it one.</p>",
+  "rule.desc.ParagraphNotUsedAsHeader": "<p>Bold and Italics are used to emphasize text, whereas headings are used to define the structure of the document. Headings like <code>h1-h6</code> are extremely useful for non-sighted users to navigate the structure of the page, and formatting a paragraph to just be big or bold, while it might visually look like a heading, does not make it one. UDOIT has flagged this as a potential usage of style for document structure, please ensure that it is not the case.</p>",
```
dmols, line 359: Little nit-picky and might be overkill, but I feel this'll make it clearer to the user what they should look for in their html. 

```suggestion
  "rule.desc.ParagraphNotUsedAsHeader": "<p>Bold and Italics are used to emphasize text, whereas headings are used to define the structure of the document. Headings like <code>h1-h6</code> are extremely useful for non-sighted users to navigate the structure of the page, and formatting a paragraph to just be big or bold, while it might visually look like a heading, does not make it one. UDOIT flagged this as a potential usage of style for document structure, please ensure that this is not the case when you review the html editor in the page.</p>",
```

dmols (changes requested):

dmols:
When you get a chance, @taheralfayad , I left some notes. Let me know if you agree or disagree.  Besides that minor input, the changes look great.

translations/en.json
dmols: I'll make that change!

dmols (approved):

PR #943, opened 2024-04-24
Update misnamed variable

taheralfayad:
Found an error relating to the modern filetype pull request that was recently merged into dev. There was a variable that was undefined that was being referenced in the `setAcceptType()` function. The original goal would have been to type in `file.fileType` rather than just `fileType`.


dmols (approved):
Replacement files no longer limited to just xlsx, docx, and pptx. Works well, clear to merge.

PR #944, opened 2024-04-24
Update nginx configurations and php.ini to increase maximum file size.

### LINKED ISSUE #942: UDOIT returns a 413 error whenever a file larger than 1MB is uploaded
UDOIT returns a 413 error if an uploaded file's size exceeds 1MB, which is due to the fact that it is using nginx's default configuration for maximum file sizes. The solution would be to manually set the file size in the `local.conf` and `deploy.conf` nginx files.

taheralfayad:
Fixes #942. In this current iteration of UDOIT, the maximum file size that the user is allowed to upload to a course is 1MB, which renders the file replacement modal useless for most PDFs. What currently happens when you try to upload a file that is >1MB is that UDOIT returns an unhandled 413 error, which isn't really that informative to the user. To combat that, I set the acceptable upload file size to 10MB in the nginx configurations for `local.conf` and `deploy.conf`. I also set `upload_max_filesize` and `post_max_size` to be 10MB in `php-custom.ini`.

### DIFF
--- a/translations/en.json
+++ b/translations/en.json
@@ -133,6 +133,7 @@
   "msg.sync.started": "Course scan started.",
   "msg.sync.failed": "Course scan failed. Course is missing.",
   "msg.file.replace.file_type": "File type not accepted. Please input a file with the correct filetype.",
+  "msg.file.replace.file_size": "File size too large. Please input a file with a size less than 10MB",
   "msg.sync.completed": "Course scan completed.",
   "msg.sync.course_inactive": "Course scan failed. Course is inactive.",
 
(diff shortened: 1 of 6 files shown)


translations/en.json
```diff
@@ -133,6 +133,7 @@
   "msg.sync.started": "Course scan started.",
   "msg.sync.failed": "Course scan failed. Course is missing.",
   "msg.file.replace.file_type": "File type not accepted. Please input a file with the correct filetype.",
+  "msg.file.replace.file_size": "File size too large. Please input a file with a size less than 10MB",
```
dmols, line 136: Nit-picky, though I feel this wording might sound better? 

```suggestion
  "msg.file.replace.file_size": "File size too large. Please input a file of a size less than 10MB",
```

dmols (approved):
Left a comment on a possible wording change, but otherwise, the changes look great! I'm able to upload files of up to <10mb and see the error modal pop up when I input a file of size >10mb. Good to push.

dmols:
Changes requested were made. Good to push!

PR #946, opened 2024-04-29
Cleanup console printouts

### LINKED ISSUE #945: Remove display of data in browser console
Might've been left there during development, but currently, when UDOIT scans a course, it prints to the console the data collected, as well as new data (if there has been more than one scan). Ideally, if we were to print something, it would make sense to just let the user know the data has been collected, whether it be new or initial data.

dmols:
Fixes #945. Also deleted other console logs that seemed to provide no insight to the user. Let me know if there are any of them you feel should probably stay or any we should also remove.


taheralfayad (approved):
Cleared console.logs as intended, and added more descriptive console.log for Data loading. LGTM!

PR #949, opened 2024-05-16
Replace File References

taheralfayad:
Presently, whenever a user attempts to replace a file using the file replacement modal, it only allows them to upload the new file to the files section and delete the previous version. However, the application does not account for references where the previous file may be used (assignments, pages, modules, etc...), and so the user ends up with a few broken links on their hands. This PR seeks to address this issue. The code in this pull request will replace all instances of the previous file in assignments, modules, wiki pages, syllabus, quizzes, and individual quiz questions, with the the new file. It also removes the deleting functionality to maintain links that the CanvasAPI does not permit us to change (announcements for example).

### DIFF
--- a/assets/js/Components/FilesModal.js
+++ b/assets/js/Components/FilesModal.js
@@ -270,6 +270,12 @@ class FilesModal extends React.Component {
       return
     }
 
+    const userConfirmed = window.confirm(this.props.t('msg.confirmation'));
+
+    if (!userConfirmed) {
+      return;
+    }
+
     let api = new Api(this.props.settings)
     api.postFile(activeFile, this.state.replaceFileObj)
       .then((responsStr) => responsStr.json())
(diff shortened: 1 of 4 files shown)


dmols (changes requested):
Not sure if this is a canvas issue or something on our end, but uploading identical files with different names or renaming files after uploading, does not translate well over to the 'Review Files' tab on UDOIT. It looks to read as if there's instances of the same file in the course, 

Screenshot 2024-07-23 at 3 50 12 PM

even though they're actually different in name: 

Screenshot 2024-07-23 at 3 50 34 PM

I also noticed that while I can upload, say a pptx file to replace a ppt file, it doesn't actually replace it in the Files page on the course. I vaguely remember this being something we talked about in the past but I don't recall if it's something we can change on our end. 

Lastly, while I can follow through with the upload on a course I create through our Dev, if I export a Canvas course of mine (need to do it like this to link a file to an assignment/module/etc) to our dev instance, the process stalls during the upload page and I get a 500 error on the console. 

Not sure why it works this way but I'm happy to help addressing these issues when you have a chance!

PR #958, opened 2024-08-22
makefile and compose improvements

### LINKED ISSUE #948: Create a Makefile
A makefile would be incredibly useful here, especially with the cache needing to be cleared every so often. Having this file would also keep us from having to dig through the repo for certain shell commands.

SimHoZebs:
## Changes
- container names are specified, ensuring no duplicate containers run under different directory names.
- all set up commands can be accessed through make commands
- database set up is a single make command instead of needing to work with the database manually

closes #948

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,4 +34,4 @@ yarn-error.log
 .phpunit.result.cache
 ###< phpunit/phpunit ###
 
-.env
+*.env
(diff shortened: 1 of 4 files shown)


dmols (approved):
Changes look good and there's no conflicts with the dev branch. Great work!

PR #978, opened 2025-01-14
8.2 to 8.4 upgrade

dmols:
This PR introduces changes that allow UDOIT to work with PHP 8.4 instead of 8.2, which no longer has active support, as of last December (read [here](https://www.php.net/supported-versions.php) for more details). 

Please let me know if issues come up running this branch. Due to the sheer volume of changes between the 8.2 branches and this one, you may have to clear the cache and run the following commands to have UDOIT load properly. 

Before running the containers: 

`rm composer.lock`
`rm -rf vendor`
`docker compose -f docker-compose.nginx.yml run composer composer clear-cache` 
and finally: 
`docker compose -f docker-compose.nginx.yml run composer composer install` to reinstall the necessary packages 

Then run the containers from scratch using: 

`docker compose -f docker-compose.nginx.yml up --build`


panbed (approved):
Great work getting the upgrade to work, I knew you could do it :-)

I've been testing the changes and it seems like the user side is good, I couldn't really find anything that didn't work at least. As far as I can see, I think we're fine to merge this into `dev`. I've also been already using these changes in the `equal-access` branch and I haven't ran into any issues either.

PR #979, opened 2025-01-15
Changed redirect uri path to fix LTI key's common issue

### LINKED ISSUE #962: Change target_link_uri in LTI config JSON
Per this page: https://github.com/ucfopen/UDOIT/blame/81b5f09014ba9136141468978a49b3d4c83e33a3/INSTALL_CANVAS.md#L23

Redirect URI should be set to 
>     * <YOUR_UDOIT_BASE_URL>/authorize/check

If this is the case, we should consider updating the controller built JSON to auto-populate this field to the correct endpoint. It currently is set to `"{$baseUrl}/dashboard"` [[code permalink]](https://github.com/ucfopen/UDOIT/blob/81b5f09014ba9136141468978a49b3d4c83e33a3/src/Controller/LtiController.php#L196).

Note: the above two `target_link_uri`s should be left as `/dashboard` and `/admin`, as they are placement specific.

dmols:
Minor change to codebase with the hope of tackling the issue where users get a 500 error while loading UDOIT, due to the fact that the LTI key changes to a default (and incorrect) value. Thank you, @dgwn for finding the error!

Fixes #962 .

### DIFF
--- a/src/Controller/LtiController.php
+++ b/src/Controller/LtiController.php
@@ -193,7 +193,7 @@ class LtiController extends AbstractController
             "public_jwk" => [],
             "description" => "User settings for UDOIT 3.x",
             "public_jwk_url" => "https://canvas.instructure.com/api/lti/security/jwks",
-            "target_link_uri" => "{$baseUrl}/dashboard",
+            "target_link_uri" => "{$baseUrl}/lti/authorize/check",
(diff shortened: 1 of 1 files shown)


dgwn (approved):
JSON now matches instructions for redirect URIs found here: https://github.com/ucfopen/UDOIT/blob/main/INSTALL_CANVAS.md

LGTM 👍

PR #984, opened 2025-01-24
React Update: 16 -> 18.3.1

mbusch3:
Based on the [React 18 Upgrade Guide](https://18.react.dev/blog/2022/03/08/react-18-upgrade-guide), I made some minor changes to the `index.js` and `App.js` files to render the React app itself. But the biggest change is that React now [recommends defining components as functions instead of classes](https://18.react.dev/reference/react/Component). This involved [adding `useState()` state variables](https://18.react.dev/reference/react/Component#migrating-a-component-with-state-from-a-class-to-a-function), [converting lifecycle hooks to `useEffect()` functions](https://18.react.dev/reference/react/Component#migrating-a-component-with-lifecycle-methods-from-a-class-to-a-function), and [destructuring props](https://18.react.dev/learn/passing-props-to-a-component#step-2-read-props-inside-the-child-component).

While I tried to maintain the vast majority of the functionality in every component, there are a few issues (like the reports not rendering properly). I figure that this isn't a real cause for concern, since most of these components will likely be rewritten in the next few months based on the new UX.


assets/js/Components/AboutModal.js
```diff
@@ -6,38 +6,30 @@ import { CloseButton } from '@instructure/ui-buttons'
...
-          <AboutPage t={this.props.t} settings={this.props.settings} />
-        </Modal.Body>
-      </Modal>
-    )
-  }
-}
-
-export default AboutModal
\ No newline at end of file
+export default function AboutModel({ t, settings, handleModal }) {
```
dmols, line 9: Should `AboutModel` be `AboutModal`?

mbusch3:
The responsiveness issues are still there. Although things seem to save. I'm still working out that, and the most recent issue is `"showExample" is not defined` when I go to revisit a previously resolved issue.

dmols:
Some things to note for future reference: 

- `Preview` image isn't loading dynamically with any UFIXIT forms where the user should be able to alter values such as HTML text, color contrast, etc. 
- When the `Save` button in a UFIXIT form is pressed, it doesn't trigger the pending state as quickly as the other branches. It still does it but there's a noticeable few seconds delay. Not sure yet what to attribute both these issues to. 

That said, they're not breaking changes and we benefit from adding this to `dev` already. Merging to the `dev-pr-testing` branch after adding a `CHANGELOG.md` entry

dmols:
> The responsiveness issues are still there. Although things seem to save. I'm still working out that, and the most recent issue is `"showExample" is not defined` when I go to revisit a previously resolved issue.

Can't seem to recreate this issue but will keep in mind for if the issue shows up on `dev`.

dmols:
> Some things to note for future reference:
> 
> * `Preview` image isn't loading dynamically with any UFIXIT forms where the user should be able to alter values such as HTML text, color contrast, etc.
> * When the `Save` button in a UFIXIT form is pressed, it doesn't trigger the pending state as quickly as the other branches. It still does it but there's a noticeable few seconds delay. Not sure yet what to attribute both these issues to.
> 
> That said, they're not breaking changes and we benefit from adding this to `dev` already. Merging to the `dev-pr-testing` branch after adding a `CHANGELOG.md` entry.

On second thought, it would be simpler to just update the `CHANGELOG` in the merged-to branch. Will make the changes there.

PR #987, opened 2025-02-06
Updating dev branch with big 4.0 release changes

dmols:
Main changes to see in this PR are as follows: 

- React upgrade from `16.14.0` to `18.3.1`
- PHP upgrade from `8.2` to `8.4`
- Equal Access scanner integration (now you can swap between `phpally` and `equalaccess`

Some things to note with this merge: 

- The preview image on the UFIXIT forms is no longer dynamic. This is a current issue, that we believe to be happening due to some of the React upgrade changes.
- When the `Save` button is pressed on the UFIXIT forms, it doesn't automatically change to a pending state; there will be a small delay. This is also likely attributed to the React change, though we're unsure as of yet.


dmols:
Met with Morgan and agreed to push these changes in, fixing the weird issues on `dev` directly, if need be.

PR #991, opened 2025-03-05
Updated github workflow 

dmols:
Updated the github workflow to avoid workflow failures due to incompatible/outdated cache versioning.


panbed (approved):
Tested on my own fork, builds correctly and cache also works. Looks good! :D

PR #992, opened 2025-03-05
New Forms (SensoryMisuse, Label, StyleMisuse) + extra bug fixes and features

panbed:
Added the SensoryMisuse, Label, and StyleMisuse form. MissingAttribute still needs to be done, but has a placeholder page for it currently.

There is also a bug fix when using phpally, where marking as resolved didn't work because I was initializing UtilityService wrong. That has been fixed.

There are some (mostly commented out) features that I've added which aren't directly related to the forms, but I already committed them to my branch, so:
- UfixitModal.js has some new code for a "data-udoit-ignore" attribute, which currently works _only_ when using Equal Access. The rule names are added to the element, and when scanning, EqualAccessService will exclude the element while scanning if one of the names in the attribute match the actual rule name from the scanner. This doesn't seem to work currently when using phpally, so it has been commented out.
- In the docker-compose.nginx.yml file, there is some commented out code for using a local Equal Access scanner endpoint in a separate container, but it needs the [server folder from Evan's equal-access-server repository](https://github.com/evannaderi/equal-access-server), so it's also been commented out.
- Finally, the LocalApiAccessibilityService has been updated to match Evan's equal-access-server endpoints and expected output.

### DIFF
--- a/assets/js/Components/Forms/LabelForm.js
+++ b/assets/js/Components/Forms/LabelForm.js
@@ -2,8 +2,6 @@ import React, { useEffect, useState } from 'react'
 import { View } from '@instructure/ui-view'
 import { TextInput } from '@instructure/ui-text-input'
 import { Button } from '@instructure/ui-buttons'
-import { IconCheckMarkLine } from '@instructure/ui-icons'
-import { Checkbox } from '@instructure/ui-checkbox'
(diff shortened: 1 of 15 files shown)


panbed:
no 💔

PR #993, opened 2025-03-06
Filter by module

mbusch3:
Added a new API call to retrieve all of the modules and their items from the Canvas API. Each issue is then given the appropriate module Id so it can be filtered by module in the FixIssues page.


dmols (approved):
Works as intended. Minor issue when it comes to multiple modules having the same inaccessible page referenced, but this will be resolved in a future PR. Merging in!

PR #995, opened 2025-03-12
Add form save function

mbusch3:


dmols:
Changed the target branch since `new-main-page-UX` is at the same stage of commits as the `new-ui-and-ux...` branch.

dmols (approved):
Found problem in UFIXIT form where resolving an issue, unresolving it, and trying to fix it as normal leaves you in a pending state where the form won't resolve. 

This may or may not be relevant to the branch itself, but noting it for future QA testing and/or branch fixes. 

Also the `AltText` form is not working as intended, so we'll be taking a closer look once more of the UI is merged in.

PR #997, opened 2025-03-17
add aria role form

nizolnier:


dmols (approved):
Good work! Functionality is there but some UI/UX changes aren't up to date on this branch. Merging with the target branch will fix it, however. Making the push!

PR #998, opened 2025-03-19
Add file to issue list

mbusch3:
Added the functionality to resolve and upload files, as well as preview and download files into their own form. They can be sorted and filtered the same as other issues.


dmols (approved):
Changes look good! Files are now able to appear in the UFIXIT forms without crashing and everything works as intended. Continuing with the push!

PR #1003, opened 2025-04-01
SensoryMisuse and Label forms

panbed:
The last merge request I made had too many random changes on it, this one _specifically_ only has the new forms. 

SensoryMisuse should be mostly finished in terms of the new UI. The other two are still using InstructureUI components.


dmols (approved):
Changes look good and the forms work as intended; thoroughly tested both forms. Only minor concern is that 'extra' might not be a sensory word, though I might be wrong on that and worst case, we can change that later. 

Note: I also did add a lot of rules for LabelForm to tackle that were missing, though I checked all of them and they work as intended. This PR is good to go, so I'm pushing this forward. Great work!

PR #1006, opened 2025-04-03
Rebuild the Admin pages without Instructure UI

mbusch3:
There are some bigger changes than just scrubbing the old UI:

In the Admin Courses, I decided that it would be very helpful to be able to select a single course for the report. So I added a "Report" button next to the "Scan" button on each Course.
- That required a new API call that got all the report data, but only from a single course instead of an entire account with many courses. I added the API route and logic to the PHP code. **That means that for this update, you'll need to clear your composer cache and do a full rebuild**

I also rebuilt a few components from the ground up because they were fully dependent on the Instructure code. The main one is `SortableTable` (and its Pagination element), but also the `ColorPicker`, which I've rebuilt with the name `ColorSelector` so I don't break the existing widget. Once the color contrast form uses the new `ColorSelector`, we can remove the old Instructure `ColorPicker`.


dmols:
Styles look good and functionality works for the reports, though it seems the filtering for the courses/users might not be working as intended. On my end, it's not pulling up any of the courses from the subaccounts, or listing the full list of users. That said, I might be going about it wrong, so it'd be helpful to play-test this with your guidance at some point @mbusch3.

PR #1007, opened 2025-04-07
Reports page new UI/UX

SimonRothUCF:
Updates the reports page to reflect the new UDOIT layout


dmols:
Changes here were added to the #1009, so there's no need to merge this in anymore. Great work, Simon!

PR #1008, opened 2025-04-08
New Homepage & Header (UX Changes)

jaadez:
Changes were made to the entire homepage to match the mockup that was designed. Components were added to show each type of error a user might receive, and a useEffect is now in place to track the errors that the report generates, as well as the user's progress. A new logo is in place, as well as a header with fresh new icons. Instructure-ui has been completely dropped, and styles are now manually written in or coming from a themes file.


dmols:
Changes here were added to the #1009 PR, so there's no need to merge this in anymore. Great work on this, Jadan!

PR #1009, opened 2025-04-10
All UI updates merged into a single branch: Plus new form styling and new ColorContrast form

mbusch3:
Sorry. It's big.


dmols (approved):
Tested thoroughly and everything works as intended. We discussed that the errors are marked as violations, despite some being recommendations and suggestions, but this will be tackled later on. We believe it's on the scanner side (Equal Access) that this issue needs to be addressed.

One other issue we noticed involves the progression bar not always working as intended; un-fixing or marking as unresolved can sometimes not be reflected accurately in the progress. Need to revisit this to find the proper cause. 

Merging this in and taking note of possible fixes for future PRs. Great work here!

PR #1010, opened 2025-04-14
Bringing all the UI changes to the forms branch 

dmols:


dmols:
Merged in CSS changes. Now we have two styles for `.btn` and `button` , which I assume affect the same components. Will work on removing one of them and pushing the change. 

More pressing is the change from `activeIssue.issue` to `activeIssue.issueData`, since the new forms are likely going to crash without that change.

PR #1011, opened 2025-04-14
Fixes for issues in the last PR for QuoteForm

barnesc3:
Two things to note with this PR:

- activeIssue may cause an error/crash
- QuoteForm may still not work with files


dmols:
Seems to be missing the check to turn a smaller quotation into a `<q>` block, but we can push this forward and make note of that in a future pass, possibly. Also need to go back and check that the preview issue gets addressed.

dmols (approved):

PR #1012, opened 2025-04-17
new css for forms

nizolnier:


dmols (approved):
Changes look good and the functionality is still working as intended! 

Making this note for later but we may want to add the functionality the new `AltText` form has, where writing in the label box causes a visual change on the preview. 

Great work on this! Merging it in.

PR #1013, opened 2025-04-21
StyleMisuse + styling and functional fixes to SensoryMisuse

panbed:
StyleMisuse is functional and the issues when trying to change fixed issues doesn't seem to necessarily be an issue with the form itself.

SensoryMisuse has added styling to the tag components so they now have a hover/active color, as well as fixes to changing activeIssue so tinyMCE will update properly when switching between issues.


dmols (approved):
Everything looks and functions well! Merging this in for the sake of QA testing. Only thing of note to put another PR for (or maybe a hotfix) would be the bug where if the text is too long and the sensory word is out of view, clicking the pill will focus on it. 

Great work on this!

PR #1015, opened 2025-04-30
Post qa pre release

mbusch3:
This adds the fix for the full rescan to rescan EVEN WHEN there are existing older report scans.

Additional minor CSS tweaks added along the way, and edited the Reports page to not show buttons that don't work yet.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1210099550935871


dmols (approved):
Changes look great! Merging in.

PR #1023, opened 2025-05-06
Added support for local equal access scanner

dmols:
These changes allow the user to use a local version of the Equal Access scanner, thus removing the need for a cloud-based solution.

Big thanks to @evannaderi, @emanuelGitCodes, and @panbed for making this possible!


dmols:
We might want to revise the leftover `printOutput` lines to see if there's still use in them or remove them altogether.

PR #1025, opened 2025-05-07
Equal access

emanuelGitCodes:
Working on adding Async to `updateCourseContent()`

Adding [Supervisors](https://symfony.com/doc/current/messenger.html#supervisor-configuration)


emanuelGitCodes:
Since I don't know the workflow behavior, I canceled it. I did not want to take the risk that this might deploy anything.

PR #1026, opened 2025-05-08
Added a "Barrier Not Found" state and notification to the content preview

mbusch3:
When the ErrorHtml cannot be found and highlighted, the user is alerted that their changes can't be saved. Still need to find a way to redraw the ufixit widget when this is the case.


mbusch3:
Second commit adds much more robust checking:

- While the contentItem is loading, the user cannot edit the form.
- After loading, if the errorHtml cannot be found, the user STILL cannot edit the form.
- In both states, all of the form fields are disabled and non-tabbable.

Also commented out the language selection tool on the Settings page. We'll put it back when there is enough Spanish translation to work properly.

dmols (approved):
Just reviewed all the changes and tested the app. All changes in `new-main-page-UX` seem to remain, and I added the changes that were lost for the `Scroll-into-view` button. Merging it in. Great stuff!

PR #1028, opened 2025-05-09
New Additions to the Fix Issues Page Filter widget

mbusch3:
- New filter available: Published Only, Unpublished Only, and Published and Unpublished
- Ability to hide or show the filters, since they now take up a full row
- New toggle in the settings menu: "Show Filters by Default"
- New clickable "Pills" to show which filters are active, which is extra helpful when the filter drop-downs are hidden

image


dmols (approved):
The changes look good and they're functional. The issues properly get fixed and saved, so it's looking better than the solution we have on `new-main-page-UX` right now. Pushing it!

PR #1030, opened 2025-05-19
Major update to saving. Takes DOM building out of PHP.

mbusch3:
Major Changes to Loading and Saving:

1) When the IBM Report returns an xpath, that xpath is now saved directly to the database as issue->html. (`ScannerService.php`)
   - This means that the server/php code no longer creates a DOM object for each issue and no longer parses out the issue xpath into html.
   - This also means that when the report object is generated, the php CANNOT check for the “ignore” classes (“phpally-ignore” and “udoit-ignore-{ruleId}”). (`EqualAccessService.php`)
   - The ignore checking is now handled on the front end each time a report is generated (`Report.js`->`analyzeReport()`)
2) When an issue and its content preview load, the issue is found directly with the xpath instead of comparing sourceHtml (`FixIssuesContentPreview`)
3) When an issue is saved or resolved, we now ALSO pass in two pieces of data: sourceHtml (because the database doesn’t have it initially) and fullHtml (this is the WHOLE new page, with the new content replacing the old).
   - Passing the fullHtml in via API means the page is processed in the front end, not the back. (`UfixitWidget.js` and not `LmsPostService.php`).
   - This requires changes to `Api.js` to accept and pass the new arguments, `IssueController.php` to read in the new `fullPageHtml` value and pass it on.
   - `LmsPostService.php` reads in the `fullPageHtml` in the `saveContentToLms` function, and uses it (FINALLY) in the `replaceContent` function.


mbusch3:
This STILL has an optimization/slowdown issue that I need to solve. Working on that next. Once it's solved, I should be able to remove a whole bunch of `console.log` statements.

dmols (changes requested):
I noticed a few things while reviewing the code that may be worth resolving: 

1. Fixing an issue works and the flip to a fixed icon is a _lot_ faster! However, there's cases where the content will be hidden in the same way the resolved issues get: 

Screenshot 2025-05-20 at 10 03 23 AM
Unsure how to recreate it, though I was also testing 'Mark as Reviewed' alongside this.

2. Marking something as reviewed hides the content altogether as well, though there's no way to mark as unreviewed in this state. 
3. If you're quick enough and mark an issue as reviewed, then unreviewed, it sends you to the list of issues in the course, except the list is not populating at all. Changing the filter brings the list back.

dmols (approved):
Changes look great! We can now save reliably, as well as mark as reviewed, unreviewed, and then reviewed again, which was an issue in the old version of UDOIT as well. 

I did see a small issue resurfacing with that for a particular 'Embedded Content Should Have Label' case but we can resolve that in a hotfix or small PR.

PR #1031, opened 2025-05-29
Updates to Existing Forms

mbusch3:
## OVERALL CHANGES:

My goal is to make the forms CLEARER and EASIER TO UNDERSTAND based on feedback from the IDs. The biggest change is that I've added an "separator" class that draws a big line between sections of the form, so that the word OR in code like this:

```<div className="separator mt-2">{t('fix.label.or')}</div>```

Becomes this:

image

I've also changed the text for quite a few labels in `en.json` to feel more like instructions.

I've added the Save button to the `FormFeedback` component so it renders consistently (spacing, disabling, etc). This means I tweaked ALL the active forms to use the new version.

### AnchorText Form:
- Added the ability to handle `<area>` elements as well as `<a>` elements, which also get flagged by the same scan rule.
- Also had to edit the `FixIssuesContentPreview` to show the `<area>` alt text, which is different than the `<a>` innerHtml.
- Added a Link Preview so users can see where the link goes and can click it to test.

### AltText Form:
- Changed text instructions and added the "OR" divider to make it clear: add alt text OR mark as decorative.

### ARIARole Form:
- Removed unused/unreferenced CSS file.
- Added information for "Clarification Text" widget.
- Added clearer instructions with the "OR" separator.

image

### Contrast Form:
- Added instruction text.
- Swapped the Text and Background editors because it felt more intuitive to have text first.

### Emphasis Form:
- Added the option to REMOVE the color styles instead of just adding bold or italic. A lot of the errors I've seen in courses are when a user has inadvertently clicked a text color, then clicks on "black" to change it back. The element now has a `style` attribute, that our scanner flags, even though it is invisible to the user.
 
image

### HeadingEmpty Form:
- Added the big fat "OR" with separator.
- If the "Delete Heading Instead" option is chosen, the text field doesn't throw an error if empty.

### HeadingStyle Form:
- Fixed an issue where if an issue with a heading level loads, it would occasionally "flicker" back and forth.
- Added the ability to select "Remove Heading Level", which replaces the `<h2>`-`<h6>` tags with a `<p>` tag.
- "Remove Additional Styling" now also removes the `style` attribute and not just internal tags like `<em>` and `<strong>`

### TableHeaders Form:
- There is now an option to mark the table as decorative, which adds the `role="presentation"` attribute. See https://www.w3.org/WAI/tutorials/tables/tips/ at the bottom of the page for the documentation on that solution.

image


dmols (changes requested):
Most of the forms work perfectly, though I do have some minor notes about the HeadingStyle form: 

- Is it possible to still highlight the heading in question when the selected heading level is 'remove heading level' ? This is what I see when I pull up the form: 

Screenshot 2025-06-12 at 8 42 16 AM

Could also make the 'Remove Heading Level' option a radio button? I'm good with either option though. It's just a little unclear what the initial error is until I start changing the heading levels. 

- It also seems to lose track of the issue as soon as you save it. 

Screenshot 2025-06-12 at 8 42 35 AM

dmols (approved):
Tested the fix and it works as intended! Pushing to `new-main-page-UX`. Great work!

PR #1033, opened 2025-06-02
Sensory Misuse hotfix: Only read words that are visible or alt text.

mbusch3:
Refines the check for sensory words to read:

Text content and alt, title, and aria-label attributes
And explicitly ignores: script, style, no script, iframe, svg, and canvas tags

We may want to add back in the iframe, svg, and canvas tags if we feel that there would be sufficient alt text there.


dmols (approved):
Code editor works well, and I no longer see keywords flagged from styling attributes. Great work!

PR #1034, opened 2025-06-05
Update for reports to fix language, make the page pretty.

mbusch3:
Fixed all of the en.json language tags so they render correctly.
Added icons for the download and print buttons, changed button placement.
Refactored where the "Export" code was being run to avoid duplication.
Removed the ReportsPage.css file, since it wasn't being used anywhere.


dmols (changes requested):
Changes look great! The buttons work as intended and I like the file name change, as well as the new text used. There's just one minor thing I noticed, that I'm hoping would be an easy fix.

When you're loading the reports page, especially the first time, the 'Print Report' button is there before the rest of the page loads: 

Screenshot 2025-06-06 at 4 04 21 PM

Is there a way to make it load when the page has been fully populated?

mbusch3:
Great catch, @dmols! Wrapped it and now it only shows after the load AND when there is data, same as the rest of the page.

dmols (approved):
The only thing I was concerned about was resolved, and now everything loads properly when the whole page populates. Good stuff!

PR #1035, opened 2025-06-09
Bringing in Features to Report Page

dmols:
This PR will bring to the reports page these additional features: 

- Print the whole report, with the graph and information tables below.
- Export to CSV option for both tables present. 

This will give instructors and easier time record-keeping, especially if they teach multiple courses.


assets/js/Components/Reports/ReportsTable.js
mbusch3, line 1: My code generally doesn't have semicolons, so this has some of Simon's older code overwriting the newer, but it's nothing that affects the table.

PR #1037, opened 2025-06-18
Quick fix: Scans check for $_ENV['VERSION_NUMBER']

mbusch3:
When a scan is run, it now saves the current version number of the code as `versionNumber` in the report "data" object. Whenever a new "sync" scan is run (when the app loads), it reads the last report's data and checks for the `versionNumber` and if it doesn't exist or is different from the current version, then instead of a targeted "only check updated content" scan, it does a full course rescan. This will help remove errors that are no longer applicable and make sure that the most up-to-date changes are consistently reflected.

Note: Once you apply this fix, it should do a full rescan the first time, since the existing report data doesn't have a `versionNumber` attribute. But you'll need to change the `VERSION_NUMBER` variable in your `.env` file to catch the difference after that.

### DIFF
--- a/assets/js/Components/App.js
+++ b/assets/js/Components/App.js
@@ -155,7 +155,8 @@ export default function App(initialData) {
   }
 
   const addMessage = (msg) => {
-    setMessages(prevMessages => [...prevMessages, msg])
+    setMessages([msg])
+    // setMessages(prevMessages => [...prevMessages, msg])
   }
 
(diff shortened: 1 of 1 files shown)


dmols (approved):
Changes work as expected! Still pending to check the `$_ENV['VERSION_NUMBER']` check but that can happen after this upcoming deploy.

PR #1040, opened 2025-07-17
Update dependencies

mbusch3:
Removed a bunch of libraries that are no longer used, and updated the ones that were. A few things to note:

- I am now ONLY using yarn, which matches the documentation. Running `yarn install` with the new `package.json` file is now sufficient and no `npm` commands are needed for any packages.
- Changes to the `tinymce-react` and `react` packages broke a lot of things. In the end, the cleanest solution was to remove `tinymce-react` altogether and use standard `tinymce` components.
- Changes to the `react-chartjs-2` and `react` packages broke a lot of things. In the end, the cleanest solution was to remove `react-chartjs-2` altogether and use standard `chart.js` components.
- Expect a significant speed up when running `make start` and when web pack recompiles code changes.


dmols (approved):
Changes look good! Node changes and `jest` updates did cause the `yarn test` step to crash, and it involves a good bit of tweaking for it to work without the workflow failing, so I'm putting a pin on resolving that for later. At the moment, we don't have any comprehensive jest testing on the app, so I commented out the step. 

Merging this in!

PR #1042, opened 2025-07-24
Final updates to UI before lock

mbusch3:
**IMPORTANT!**
Among these commits, `4bb4094` adds a new route to the API. When you recompile, you'll need to delete your `var/cache/prod` folder or the routes may not recompile and you'll get a 404 error when you scan the course.

Updated text across multiple forms to clarify the "How to Repair This Barrier" instructions.
Moved around multiple things to clarify the Sensory Misuse form.
Also fixed File form to enable "Mark as Reviewed", which was coming up as disabled.


dmols (approved):
Changes look good and the styles fit well. Resizing the window also works well with the styling change. All the forms work as intended and the new settings are great (May have to add a few tweaks in the future for the dark mode)! Pushing this to `new-main-page-UX`. 

Thanks for the great work!

PR #1048, opened 2025-08-27
Rotate keys functionality and removing hardcoded encryption key

panbed:
im praying it works


dmols (approved):
Changes look good! I imagine the `INSTALL.md` will eventually get updated later by #980. This one should take priority, since `main` needs it.

dmols:
Testing this fully requires that I brick my database, so I'll test some other PRs first and circle back to this one!

PR #1053, opened 2025-09-16
Adds tooltips to issues table

### LINKED ISSUE #1052: Add issue quick view in reports issues table
ENHANCEMENT
Add button for each entry on table to see a pop up with a description of the issue, so you don't have to go to the barriers tab and click through some interfaces just to see what the error is about (and then click back on reports to go back to the issues table).

ewainberg:
This PR adds a button next to the issues table of the reports to quickly see a barrier's description.
Fixes #1052


assets/js/Components/SortableTable.css
mbusch3, line 1: By moving the `clickable` class to a cell (<td>) instead of the row (<tr>), the hover effect becomes kind of weird: only the left-most cell changes color, when the whole thing probably should.

assets/js/Components/Reports/IssuesTable.js
```diff
@@ -116,13 +189,66 @@ export default function IssuesTable({
...
+      <SortableTable
+        caption={t('report.title.issues_by_type')}
+        headers={headers}
+        rows={rows}
+        tableSettings={tableSettings}
+        handleTableSettings={handleTableSettings}
+        t={t}
+      />
+      {/* ARIA live region for screen readers */}
+      <div aria-live="polite" style={{position: 'absolute', left: '-9999px', height: 0, width: 0, overflow: 'hidden'}}>
```
mbusch3, line 202: I love that you are using the correct accessible semantics: aria-live="polite" to make sure that the screen reader text interrupts, but not until the previous text is done, and role="dialog" below.

Is there a compelling reason to use a div with the role of dialog instead of a dialog element?

assets/js/Components/Reports/IssuesTable.js
```diff
@@ -116,13 +189,66 @@ export default function IssuesTable({
...
+        {ariaLive}
+      </div>
+      {popover.open && (
+        <div
+          id="issue-info-popover"
+          role="dialog"
+          aria-modal="true"
+          tabIndex={-1}
+          ref={popoverRef}
+          style={{
```
mbusch3, line 212: For permanent styles, try to keep it in the CSS. I'm assuming that you did this because of the `popover.x` and `popover.y` values, which IS the right way to do this in React. Another approach could be to use CSS's attr() function. So give the element a dummy attribute, like `data-pos-x={popover.x}` and `data-pos-y={popover.y}`, then in the CSS, say `left: attr(data-pos-x); top: attr(data-pos-y)`.

assets/js/Components/Reports/IssuesTable.js
```diff
@@ -116,13 +189,66 @@ export default function IssuesTable({
...
+          <div
+            style={{ marginBottom: 8 }}
+            tabIndex={-1}
+            ref={popoverContentRef}
+          >
+            {popover.content}
+          </div>
+          <button
+            type="button"
+            style={{
```
mbusch3, line 238: This styling should DEFINITELY be in CSS, since there isn't anything dynamic about it.

assets/js/Components/Reports/IssuesTable.js
```diff
@@ -116,13 +189,66 @@ export default function IssuesTable({
...
+            minWidth: 220,
+            maxWidth: 320,
+            fontSize: "0.95em",
+            outline: "none",
+          }}
+          onClick={e => e.stopPropagation()}
+        >
+          <div
+            style={{ marginBottom: 8 }}
+            tabIndex={-1}
```
mbusch3, line 231: I have NO IDEA why, but each time I use the keyboard and the popup turns on, the focus goes straight to the text, even though it's explicitly set to `tabIndex='-1'`. Not sure how to approach that one.

PR #1055, opened 2025-09-18
Fixes crash when there is no background color set

### LINKED ISSUE #1054: Crash: Color Contrast Form when no background color set
Crashes when there is no background color set in the contrast form

ewainberg:
Fixes #1054


ewainberg:
Fixes #1054

mbusch3 (approved):
Thanks, Elias!

PR #1056, opened 2025-09-18
Adds user name display on Users table in Admin reports

### LINKED ISSUE #476: Admin reports > Users table shows user ID, but not their name
ENHANCEMENT IDEA
On the other Admin page tabs, when it shows the numerical IDs it pulls down the course names down in realtime and overwrites the number (e.g. on the Scans report the Term, Course, User columns). It would be great if the User column on the Users report pulls the name down too.

ewainberg:
- Pulls user's name from canvas API on users table load
Fixes #476

### DIFF
--- a/src/Controller/AdminController.php
+++ b/src/Controller/AdminController.php
@@ -15,6 +15,7 @@ use App\Services\UtilityService;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Routing\Annotation\Route;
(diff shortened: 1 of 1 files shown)


mbusch3 (changes requested):
While this absolutely does the trick (I've tested it locally and it works well), I think we could do it a better. Is there a way use this logic elsewhere (like when creating the user), and save the user's name to the user table.

I'm concerned that EVERY time you load the page, you'd need to do the API call for EVERY name, and sometimes the lists can be very long.

PR #1058, opened 2025-09-19
Show instructors in Admin Reports and populate user full names in DB

barnesc3:
# Summary

### Users

- Full names are now stored at creation (`User.name` is populated; no more nulls).

### Admin Reports

- **Courses Page**: Added an **Instructors** column in the table.
- **Single Course Report**: Instructor list displayed under the course title (renders only when present).
- **Refresh Logic**: Instructor list refreshes from the LMS when stale (TTL = 24h).

---

# Key Changes

## Configuration

- LTI tool config updated to provide the user’s display name (`privacy_level = name_only`).
- Existing session → DB logic now ensures `User.name` is set during user creation.

## Backend

- Introduced **`course_user`** join table (`Course` ↔ `User` mapping) to cache instructors:
    - **Columns**: `course_id` (FK), `user_id` (nullable FK), `lms_user_id`, `display_name`, `fetched_at`.
    - **Constraints/Indexes**: Unique (`course_id`, `lms_user_id`); indexed on `course_id`, `user_id`.
- LMS integration fetches active teacher enrollments (fallback: users) and upserts into `course_user`.
- API now returns `instructors: string[]` for:
    - `GET /api/admin/courses/account/{account}/term/{term}`
    - `GET /api/admin/courses/{course}/reports/full`
- Cache invalidation: refresh from LMS if empty or older than 24h.

## Frontend

- **CoursesPage**: Added **Instructors** header/column with comma-separated names.
- **ReportsPage (single course)**: Render instructor list under the title; hidden when empty.

### DIFF
--- a/INSTALL_CANVAS.md
+++ b/INSTALL_CANVAS.md
@@ -46,6 +46,7 @@ We strongly recommend you enforce scopes with your API key. The following scopes
   * url:PUT|/api/v1/courses/:id
   * url:GET|/api/v1/courses/:id
   * url:POST|/api/v1/courses/:course_id/files
+  * url:GET|/api/v1/courses/:course_id/users
 * discussion_topics
   * url:GET|/api/v1/courses/:course_id/discussion_topics
   * url:PUT|/api/v1/courses/:course_id/discussion_topics/:topic_id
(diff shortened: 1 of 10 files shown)


mbusch3:
This is **wonderful** work and seems to be working for the most part.

I have a couple issues that I'm seeing when I run it on my end:

- When a course has not been scanned, I'm usually not getting people's full names. For instance, a course on my sub-account listed the instructors as "Corey, Erika" before a scan, and afterwards as "Corey Peterson, Erika Alexander." I am not sure why it shows differently.
- When I try scanning multiple courses at the same time, I get odd results where some things come back and others just hang. One solution would be to just disable all of the scan buttons when any course is scanning, and another would be to make multiple synchronous scans work consistently.

That said, I haven't found any other major issues, and the report screen updates data immediately. Good work and well done!

dmols (approved):
Changes look good and the admin page works as intended! Great work!

PR #1061, opened 2025-09-29
Sensory misuse no word bug

Ishfaq-code:
There were instances where the `SensoryMisuse` issue was finding text with no sensory words. Issue was much different where it was finding it but the error element pointed to by the xpath was different. Turns out our equal access scanner was not saving xpath with a `main`. 

Fix:
Don't check for `main` or `div[1]` in `html.js`


dmols:
Closing this PR in favor of this other one (#1065), since this one contains code not needed for the fix.

PR #1062, opened 2025-10-01
Finished text summaries and descriptions, adds ability to link text from other issues when it's repeated.

mbusch3:
# RULE DESCRIPTIONS IN UDOIT

This PR supersedes and replaces #1045, which I am closing. In addition to completing all of the English-language summaries and descriptions for individual Equal Access ruleIds, this commit also standardizes and cleans up how the text is organized in `en.json` and referenced elsewhere.

## GETTING TEXT OUT OF FORMS AND RULES

This PR contains written text summaries and descriptions for every IBM ruleId encountered in UDOIT. Here's how it works:

### CASE 1: PART OF A FORM
If a rule is handled by a form, then that form's `title`, `summary`, and `learn_more` sections are used to teach about the issue. If additional information would be helpful for fixing the issue, that is included in the rule's `info`. For ease of use, I've included all of the blank rule descriptions with the forms that the rules are associated with. The rule descriptions are blank because whenever the user encounters this rule, it's using the form.

For example:
```
"form.aria_role.title": "HTML Element has an Incompatible ARIA Role",
"form.aria_role.summary": "ARIA roles are often used to provide needed context for screen readers, but many HTML elements have built-in roles that sometimes conflict.",
"form.aria_role.learn_more": "<h3>ARIA Roles</h3><p>ARIA (Accessible Rich Internet Applications) roles ..."
...
"rule.desc.aria_role_allowed": "",
"rule.desc.aria_role_valid": "",
"rule.info.aria_role_valid": "This HTML element is a/an <{tagName}>, but has an ARIA role of \"{ariaRole}\", which is incompatible.",
```

### CASE 2: NOT PART OF A FORM, BUT STANDS ALONE

If there is NOT a form to handle the rule, then the ruleId has its own `summary` and `description`. The `description` appears in the `learn_more` area, and is analogous to the form's `learn_more` text.

For example:
```
"rule.summary.text_whitespace_valid": "Ensure that spaces and blank characters are not used to create space between letters of a word.",
"rule.desc.text_whitespace_valid": "<h3>Consistent Spacing</h3><p>Adding extra spaces or blank characters..."
```

### CASE 3: NOT PART OF A FORM, BUT INHERITS TEXT

Often, multiple rules can use the same `summary` and/or `description` as other rules. When this is the case, a rule may not have an entry for one or both of these. In these cases, they are using text supplied for another issue. In this case, the ruleId can reference any other index in the translation file. The references are handled in the `Ufixit.js` file:

```
export const sharedRuleSummaries = {
  blink_elem_deprecated: 'rule.summary.blink_css_review',
  ...
  style_viewport_resizable: 'rule.summary.meta_viewport_zoomable',
}
```

and

```
export const sharedRuleDescriptions = {
  aria_banner_single: 'form.label.learn_more',
  ...
  widget_tabbable_single: 'rule.desc.widget_tabbable_exists',
}
```

## OTHER CHANGES

- Added another "Affected User" type (foreign language speakers)


dmols (approved):
Changes look good! Great to see all the translations in the app already! Some changes in the `es.json` I may need to change back, as they're small changes I added to `new-main-page-UX` a while back that didn't persist here.

PR #1063, opened 2025-10-01
Several small bug fixes...

mbusch3:
Each commit is a small, self-contained bug fix:

[Addresses failure to save when element is deleted.](https://github.com/ucfopen/UDOIT/commit/d4f48a34f664cbe6ffe37cf53e6103eb2a997ca3)
[Adds specific rule conditions for the AltText form.](https://github.com/ucfopen/UDOIT/commit/8b80c5053c29c9a546047ab068b093ce649f4b7a)
[Fixes issue where combobox display:block overrode flexbox settings.](https://github.com/ucfopen/UDOIT/commit/ead09d7420820278494fb0ec68900bfb81b6f46b)
[Fixes a really stupid issue where undefined !== null and it led to a crash.](https://github.com/ucfopen/UDOIT/commit/43a5b892adea36c172ce842efe419aa92f87db75) 
[Adds additional cleanup for table_structure_misuse errors.](https://github.com/ucfopen/UDOIT/commit/4e6d661f82e43f804aa8f80effc73b8daf376599)
[Removes unused devDependency.](https://github.com/ucfopen/UDOIT/commit/50537836f6b96df1da050562ce2a6c0c914e373b)

### DIFF
--- a/assets/js/Components/FixIssuesPage.js
+++ b/assets/js/Components/FixIssuesPage.js
@@ -463,7 +463,10 @@ export default function FixIssuesPage({
 
     setFilteredIssues(tempFilteredContent)
     setGroupedList(groupList(tempFilteredContent))
-    setActiveIssue(holdoverActiveIssue)
+    if(holdoverActiveIssue) {
+      setActiveIssue(holdoverActiveIssue)
+    }
+    
   }, [report])
 
 
@@ -702,7 +705,7 @@ export default function FixIssuesPage({
     return tempDoc.body.innerHTML
   }
 
-  const handleIssueSave = (issue, markAsReviewed = false) => {
+  const handleIssueSave = async (issue, markAsReviewed = false) => {
 
     if(!activeContentItem || !activeContentItem?.body || !issue) {
       return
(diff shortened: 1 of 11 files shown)


dmols (changes requested):
Not particularly sure how to test the `<canvas>` and `<input>` specific condition changes for the `AltText` form, seeing as both are not allowed in Canvas. Could do so in an html file but since the file replacement feature is not fleshed out (iirc), it may be difficult to test if it works. Open to suggestions!

That said,  I did test the rest of the issues fixed and I saw no errors pop up except for the "Heading Empty" and "Links Should be Descriptive" fix. This is what I saw when I tried to remove a heading. 


https://github.com/user-attachments/assets/2be98b1a-8e88-4cd9-9131-809664e090c1

Same thing occurs for the links, though sometimes clicking "Show Content Preview" just loads a page with the word 'undefined' in it, like so:

Screenshot 2025-10-07 at 9 59 07 AM


Worth noting, I did clear my code cache and browser cache, and still encountered the same issue.

mbusch3:
@dmols Good catch there. This is a weird one.

So, it turns out that the local "activeContentItem" is getting saved with the new, updated version of the content BEFORE the save function is returning from the API. So if there is an issue, and the save DOESN'T complete for any reason, there will be a discrepancy between what is in the local version and what is in Canvas. A full course rescan fixes it, but that's obviously not the desired behavior.

The SECOND issue is that when you choose to delete the Heading in your example above, it really, really should not be throwing that "No changes, so I can't save" exception. I'm going to look into that first and the above issue next. Hope to have a fix today or tomorrow so we can get this PR wrapped up.

dmols (approved):
Most recent changes look great, and the rest is working as intended still. Good work!

PR #1064, opened 2025-10-01
Language Form Final

Ishfaq-code:
Added language form


dmols (changes requested):
Works well and does what it needs to do, though there are some places where it could use minor improvements. 

I left some comments already for potential changes in the code, but one minor functional thing I noticed is that when we have several issues that the language form picks up, if I fix one and move onto the other, there's the potential to show errors when visually everything looks fine. 

Screenshot 2025-10-03 at 9 54 05 AM

The solution would probably be to reset the values for each input when `activeIssue` changes.

PR #1065, opened 2025-10-02
SensoryMisuse bug Fix

Ishfaq-code:
There were instances where the `SensoryMisuse` issue was finding text with no sensory words. Issue was much different where it was finding it but the error element pointed to by the xpath was different. Turns out our equal access scanner was not saving xpath with a `main`. 

Fix:
Don't check for `main` or `div[1]` in `html.js`


dmols (approved):
Co-tested this with Ishfaq and we saw it remove the cases where we'd have the form pop up saying there were no sensory words found. Good to go!

PR #1069, opened 2025-10-16
Adds gradients and related features, also fixes contrast form parent issue

ewainberg:


dmols (changes requested):
The form works great! I just have a few requests that I believe would make the form a little better: 

- I would make it so that by default it only shows 3-4 colors in the instructions but allow the user to show the full list if there's more than that. This way, if the user is just trying to navigate past the issue, the 'next' button will be visible and not pushed down out of frame. 
- Navigation is nearly done, I believe. However, tabbing through the colors should probably associate the 'X' or check mark to the color, when focusing on the color div, so an unsighted user can know the color isn't valid yet.

dmols:
@ewainberg I met with Morgan, and we both agreed the alert for both the foreground and background color should be something like "Insufficient contrast of [current contrast number]. Desired contrast is [correct contrast number]" or something shorter, if you're able to condense it further without losing information. All the information you need for the correct contrast number, your code logic already handles, so you shouldn't have too much trouble giving that value to your alert. 

Let me know if you have any questions! Thank you in advance for adding this change!

assets/js/Components/Forms/ContrastForm.js
```diff
@@ -428,12 +418,21 @@ export default function ContrastForm({
...
               </div>
               <div className="flex-column justify-content-center">
-                <div className="ratio-value">{contrastRatio}</div>
+                <div className="ratio-value">{contrastRatio?.toFixed(2)}</div>
               </div>
             </div>
             <div className="flex-row justify-content-center">
-              <div className={`ratio-status ${ratioIsValid ? 'valid' : 'invalid'}`}>
-                {ratioIsValid ? t('form.contrast.feedback.valid') : t('form.contrast.feedback.invalid')}
+              <div className={`ratio-status text-center ${ratioIsValid ? 'valid' : 'invalid'}`}>
```
ewainberg: Forgot to mention it in the commit but this block shows the minimum ratio needed in the pass / fail box

PR #1070, opened 2025-10-20
Adds Quiz Question saving and scanning.

mbusch3:
This PR was initially started because I was trying to figure out the stupid "inconsistent scan results" bug. I started by cleaning up a LOT of old PHP that isn't used anymore and trying to make sure that consistent functions are called regardless of whether content is scanned from the full rescan or the quick checkup scan.

I've changed the `.env` variables to remove the lambda references and list the excluded rule Ids. The `.env.example` file is updated to reflect that.

The biggest change is that now Quiz Questions can be scanned, show up on the report (under the "Quiz" content type), and can be saved back to Canvas properly. That one wasn't really all that many lines of code all told, but was a pain to get working. The biggest change is that we are now saving an "parentLmsId" in the content_item metadata. The saving and updating functions now use that to create the proper URL (/courses/course_id/quizzes/QUIZ_ID/questions/question_id).


dmols (changes requested):
Works as intended! I only have a few notes of potential changes we can add to make the experience a little better: 

- Unlike the changes made to other forms of content on a course, it seems changing anything inaccessible in a quiz puts it in a sort of 'draft' state. This means the faculty member needs to go back into the quiz and save the changes. 
Screenshot 2025-10-24 at 7 51 31 AM

I would propose that we make that known somewhere in the form instructions (maybe as an additional info note based on if the form refers to quiz content?) 

- Less important than the one above, but if the accessibility issue involves a quiz question, clicking the arrow to go to the page will take the user to the main page of the quiz, not the quiz question that has the issue. Is it possible to redirect users to the quiz question instead of the main quiz? Not too annoying for small quizzes, but I could see this being a little annoying for big quizzes.

dmols (changes requested):
Functionality is still good from last time, and I like the most recent additions! I think the only thing that needs addressing is the fact that images will flicker on this branch. It's strange that this is happening, seeing as the fix from this PR #1029 is in the target branch `new-main-page-UX` already. I've attempted to bring over some of the changes from that PR in hopes of fixing the issue, but it seems to just make the behavior more aggressive, even when there aren't problems compiling. 

Once the flicker weirdness is addressed, I believe we can push this PR forward!

dmols (approved):
Talked about it and the fix for the flicker didn't seem to actually solve the issue, so it's possible we're stuck with the flicker, since it's a React issue. We'll look into it, but it's not breaking anything, and the code functionality here will be useful in `dev` and `main`. Merging on!

PR #1073, opened 2025-10-28
Adds an async batch fetch function for getting Canvas Pages.

mbusch3:
Added a new `apiGetBatch` function to `CanvasApi.php`. It allows you to enter an array of URLs that need to be fetched and uses the `curl_multi_*` functions to create and watch all the handles and return once they've all gotten responses.

I've applied it to the `updateCourseContent` function in `CanvasLms.php` for pulling all the Page content from Canvas. On my own course that has 43 pages that need to be fetched, running a Full Course Rescan gave me the following times (in seconds):

| Method | Test 1 | Test 2 | Test 3 | Test 4 | Test 5 | Test 6 | Average |
| --- | --- | --- | --- | --- | --- | --- | --- |
| Old/Sequential | 9.504 | 10.286 | 9.633 | 9.914 | 9.529 | 10.185 | **9.842** |
| New/Async | 6.540 | 6.689 | 6.895 | 6.734 | 6.623 | 6.584 | **6.678** |

Obviously, the number and size of course Pages may drastically change your personal results.

I've kept all the testing code (including timers and readout) and both fetch methods so you can test it on different courses yourself. Just change the `$asyncFetch = true;` to false in the `updateCourseContent` function.


dmols (approved):
Changes look good, and we have no issues in the pipeline. Merging through!

PR #1084, opened 2025-12-01
Updated message tray

mbusch3:
Before, only one alert message could be displayed, and in another alert came through, the first alert would be overwritten. Now, multiple alerts can display on screen, each with its own timer and "close" button.

The easiest way to test this out is to change multiple settings on the Settings page or open another tab and switch between the UDOIT tab and the other a few times.

image

Code-wise, this completely reworks the MessageTray component and adds the Message component for an individual message. The App.js code is edited to just send each message as it receives it directly to the MessageTray, where it is handled, instead of having a setMessages function at the App level.

### DIFF
--- a/assets/css/udoit4-theme.css
+++ b/assets/css/udoit4-theme.css
@@ -69,6 +69,8 @@ body {
 }
 
 #app-container {
+  position: relative;
+  overflow: hidden;
   margin: 0px 6px 6px 0px;
   padding: 0 1rem;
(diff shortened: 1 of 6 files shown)


dmols (approved):
It might get a little busy if you keep cycling through tabs but that's an edge case. With this we'll be able to easily see different alerts on the screen at once, which I think is a good addition! Code works well, pushing to `dev`. Good work!

PR #1082, opened 2025-11-19
Admin reporting dashboard

barnesc3:

PR #1086, opened 2025-12-03
Admin reporting dashboard

mbusch3:
Replaces PR #1082. There is still work to do with regards to final choice of data (How are barriers listed? Suggestions? Files scanned vs reviewed?) as well as layout.


dmols (approved):
Changes look good and function well! The interface is also very sleek! I also do like that we now have a separation between the dashboard and the `courses` tab. Great work!

PR #1093, opened 2026-01-08
Updated Barriers page

mbusch3:
- Added page title and subtitle (now matches file page and mockups)
- Updated filters to have toggles for showing resolved and unpublished barriers. That means when you come to the page, all the filters are OFF be default, and can be added as desired.


Ishfaq-code (changes requested):
Reviewed forms from `AltText.js` - `EmbeddedTileForm.js`, as well as the barrier page and reports page (although this may not have been included in this PR).

### Barriers Page
On the barriers page when listing the issues it shows up with both the content item title and below it the module it is located in. Some of the content item are not in a module which makes the style look a bit odd.

Image

Other than that the barrier page looks good visually and seems keyboard and screen reader accessible 

### Forms
All the forms seemed functional, keyboard accessible and screen reader accessible. There were some inconsistent behaviour throughout:
- On the `AltText.js` form, sometimes the options come preselected which seems fine for now but is consistent which may create confusion. Might be better to fill in all values in the form without selecting them so when a user does select an option they see the values associated with it.
- On some of the forms after removing the targeted element (such as removing link in the `AnchorTag.js` or marking an image as decorative on `AltText.js`) you are able to to 'remove' them again (not really removing them but are able to perform the operation again). Wondering if it might be possible to disable the option or not show it if a link is not present, or an image already decorative

### Report Page
Although this may not be the focus of this PR, did notice a bug with the page when using keyboard. Tabbing through the barriers table and going onto one of the barriers and focusing on the question mark (the SVG/Icon) leads to opening the error on the barriers page rather than opening the `LearnMore` modal. Image below identifies the icon that behaves this way:

Image

Ishfaq-code (approved):
LGTM!

dmols (approved):
Changes look good and we're clear to merge!

panbed (approved):
Looks good!

PR #1096, opened 2026-01-15
Removes code blocks for Lambda

mbusch3:
Dirty fix to make the equalAccess scanner always do the same thing regardless of equalAccess scanner ENV variable.


dmols (approved):
This change looks good, and I think it should help us not end up using lambda again.

PR #1098, opened 2026-02-04
New Summary Screen and Navigation Header

mbusch3:
Fully accessible and responsive summary screen and navigation header!

image

- Summary Screen is based on the new Deutsch (tm) designs, much cleaner and simpler.
- Removed several of the circle dials
- Added simpler linear progress indicators
- The "most important thing" is subtly shaded with a gradient background to let you know what you're supposed to do at the moment.
- Fully mobile friendly, with rows that wrap as the screen gets narrower.

image

- Navigation is now accessible and doesn't announce unnecessary groups and icons.
- Mobile drop-down nav toggle appears on narrow screens.
- Mobile navigation disappears when an item is selected.


Ishfaq-code (changes requested):
Reviewed the Summary Page + Navigation header as the PR addressed. Everything seems to be functional and accessible. Keyboard navigation is responsive and the user is able to tab through interactive elements. Screen reader is also working and reading out everything it needs.

Though there is a bit of an issue (maybe intentional) when navigating the Summary Screen using the screen reader. Whenever the screen reader is reading a barrier description it breaks up into three parts and the user has to manually move to the next part for the reader to read it out. It feels a bit weird being disjoint as such and even weirder if the user is navigating backwards and the screen reader reads the last part first as shown in the video below:

https://github.com/user-attachments/assets/71b521c1-c7b8-4d05-bba5-12c3b7252179


Otherwise LGTM

Ishfaq-code (approved):
LGTM!

dmols (approved):
UI looks good and the functionality is still as expected! Great work!

panbed (approved):
Looks good!

PR #1100, opened 2026-02-09
New settings page

mbusch3:
Updated the Settings page... and a few other things.

- Settings page mostly matches the Tory(tm) layout and is now fully responsive.
    - Realized we aren't using the Lexend font to its full potential. Imported TrueType font to start that process. 
- Realized the mobile header didn't have a good breakpoint and made that dynamic, too.
    - There's now a JavaScript way to check if the horizontal nav will fit its container and turn on the mobile nav.
- Playing with the CSS made me realize that layers weren't working because our imports were a mess.
    - Long story short, the only reason the theme was being included was that it was imported by another CSS file, which was only accidentally being loaded anyway. Cleaned up that code.
- Updated the en.json and es.json files with new Settings page text.
    - And then added a bunch of missing lines to the es.json text file. They now match.


Ishfaq-code (changes requested):
The page appears nice and responsive. It is navigable using the keyboard and supports screen reader use however there are couple of slight issues.

1.  When using a keyboard and the screen reader to select or change options in the comboxes in settings, the first input is ignored and combobox opens on the next input. What I am assuming is happening is that the combobox selects the selected option again before opening it. This is also the case with the dark mode toggle where it doesn't toggle on the first enter. This also triggers the 'settings changed' alert even though nothing is technically changed (this only seems to happen when navigating using Caps Lock and arrow keys and not during tab)

https://github.com/user-attachments/assets/c6bd3ba7-bf12-4931-bf8b-576b6e346a24

2. There are no links for the `Release Notes`, `Documentation`, `About UDOIT`, and `About UCF Open` hyperlinks and when clicking on them they just take you back to the course scanning screen

3. You can spam changing settings which then spams the alert which takes up the screen, slightly lags UDOIT and messes with the timer of the alert (some alerts timer gets reseted when one of the previous alert finishes) .

### **EPILEPSY WARNING**

https://github.com/user-attachments/assets/dfe243d0-5aba-4884-becb-ee523622982a

Ishfaq-code (approved):
LGTM!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

PR #1108, opened 2026-02-12
Rebuilt the reports page

mbusch3:
Rebuilt the Reports page based on the new UI. While I was at it...

- Fixed some spacing issues on the Home Page (summary screen) and the Settings Page.
- Created a new type of widget called a SliderSelect, which lets the user pick from a number of options in a slider instead of a dropdown.
- Fixed the report printing (text for labels instead of [Object object] errors)
- Fixed the CSV data export function (same type of [Object object] errors)
- Fixed the report grouping by form type
- Fixed some issues with the InfoPopover where the focus would pull when it was drawn, and where activating the popup with the keyboard always drew the widget in the corner of the screen.
- Added new text in English and placeholder text in Spanish for multiple new phrases and labels.

image


dmols (approved):
Changes look good and functionality is as expected!

PR #1109, opened 2026-02-13
Welcome screen UI refresh

mbusch3:
Rewrote the Welcome screen to match the new UI.

- Has a cute progress animation bar.
- Realized that the messageTray wasn't showing up and fixed that bug, too.

image


dmols (approved):
Merging in `dev` fixed the weird issue where the scan would timeout slightly. Found some deprecation errors but most of them are not really solvable due to having vendor packages and no access to some of their specific functions. Will likely get rid of some of these anyways in future upgrades. Good to push, great work!

PR #1114, opened 2026-02-23
Fix for page batch API calls being slower than sequential calls 

Ishfaq-code:
When async fetch enabled, it would put all pages up for scanning rather than only putting updated pages for the scan. 

Now checks if a page is updated by the user before putting it in the list to fetch in batch and scan.

### DIFF
--- a/src/Lms/Canvas/CanvasLms.php
+++ b/src/Lms/Canvas/CanvasLms.php
@@ -364,9 +364,6 @@ class CanvasLms implements LmsInterface {
         // Batch page pulling maintenance
         $pageUrls = [];
 
-        // Due to the sequential calls, this field being true currently makes the process a lot slower.
-        $asyncFetch = false;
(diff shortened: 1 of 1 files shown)


Ishfaq-code:
> Not to be addressed here because I don't think the root of it is in these changes, but I'm still seeing the timeout error when running full course rescan. Worth noting for future `dev` testing. 
> 
> As far as the changes here go, the slowdown we used to see before is gone! The batching seems to be working properly now. Asides from the minor comment below, this PR is good to merge soon! Good work fixing this! 



The reason for the timeout is probably due to the fact that it switches between sequential calls and batching and there are some checks in between to do so (almost taking extra steps). It might be better to rewrite the function to just do batching.

dmols:
Closing this PR since batching API calls is something the current state of UDOIT can't handle.

PR #1117, opened 2026-02-24
Delete unused files

Ishfaq-code:
### Summary 

Adds a button to the `Review Files` page that allows for files to be deleted in bulk. Users are able to select all files or individually select files to delete

The deletion is done by batch and takes in a payload of files to be deleted once at a time. This is done to avoid hitting the limit on the canvas API. By default 10 files can be deleted at a time. This means if there are 22 unused files in a course, 10 files go get deleted, then the next 10 and then finally the last 2

### Testing

1. Upload files onto canvas but don't use them on pages. This can be done by using the files tab on a canvas course
2. Go onto UDOIT Review File Tab
3. Click the `Delete Unused Files` button on the top right
4. Option to delete should open up 
5. User can toggle to select all or select 1 by 1
6. Click the delete button on the bottom right
7. After all files are deleted the modal should update on show that
8. The delete button should be disabled when the files are in the middle of being deleted

### Feedback

Functionally the implementation was working so I am looking mostly for UI/flow feedback. This is a significant feature that maybe used a lot and making the flow simple, easy to understand and clean is important.


mbusch3 (approved):
Good to go! Minor changes due to another PR have been addressed.

PR #1122, opened 2026-03-02
Rework for the Barriers and Files full-screen modal popups.

mbusch3:
Before, when selecting an item the Barriers and Files screens, there was a full-screen `<dialog>` element where fixing all the issues happened. But using a `<dialog>` caused a number of issues:

1. Any other element in the main area was automatically hidden underneath the dialog. This meant that the notification popup (`MessageTray.js`) became invisible. So all of the "Content has been successfully saved." or "Error: ..." messages weren't available exactly when they needed to be.
2. OTHER dialogs didn't work inside them. Two notable examples are the `InfoPopover.js` element (Elias' `i` icon with information that shows up afterward) and the code editor widget inside of the TinyMCE editor.

So I've made the full screen dialog just a `<div>`, but with the right aria attributes to make screen readers THINK that it's a normal `<dialog>`. As far as I can tell, all normal keyboard, screen reader, and mouse interactions still work the same.

### DIFF
--- a/assets/css/udoit4-theme.css
+++ b/assets/css/udoit4-theme.css
@@ -536,8 +536,8 @@
     outline: 2px solid var(--focus-color);
   }
 
-
-  dialog {
+  dialog,
+  div[role="dialog"] {
     width: 90vw;
     max-width: 50em;
     overflow: hidden;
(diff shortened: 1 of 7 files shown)


dmols (approved):
Checked out the branch and the message popup shows before the dialog as intended. Screen reader is also working the way we want to. Changes look good!

PR #1127, opened 2026-03-05
Cleans up authentication process and removes the toolSettings div

mbusch3:
There is a minor security issue during the login process, where a LOT of data for user authentication and settings is being passed to the app by creating a `<script>` div with the id of `toolSettings` that contains a JSON object with everything from the user's preferences to their auth_token.

## Preferences and ToolSettings Cleanup

The old process looked like this:

- PHP in the DashboardController renders a stub of a page using the `index.html.twig`. It stores important information in the `toolSettings` script.
- In the index.js file, it runs the `getInitialData` function to read in the same JSON data from `toolSettings` to pass it into the React App.
- React App launches with the right dark mode, language, and other user settings without any problems or flickering.

The new process looks like this:

- PHP in the DashboardController renders the entire page, including the build process's CSS and JS file references.
- In the index.js file, the `getInitalData` is called as an API call with the settings data returned.
- AFTER it returns, the app is loaded with that data. (`getInitialData('api/settings').then((data) => {`)

There is a LOT of cleanup that was required for this to work:

- There is are two new routes: `api/settings` in the DashboardController.php file and `api/admin/settings` in the AdminController.php file.
- The getInitialData.js file calls these endpoints (with an async function)
- All of the .html.twig files have been removed and replaced with direct HTML responses from the relevant PHP files.

## Auth_token Cleanup

Even then, we DON'T want to send the `auth_token` as a setting. Instead, that has been stored in the PHP Session and response headers using the `use Symfony\Component\HttpFoundation\Cookie` library.

- In `DashboardController.php`, `AuthController.php`, and `LtiController.php`, whenever the auth_token USED to be passed (and exposed) as a GET parameter, instead, they are explicitly embedded in the header. For example, in `LtiController.php`, line 111 / 120:
```
$redirectUrl = $token->{'https://purl.imsglobal.org/spec/lti/claim/target_link_uri'} . '?auth_token=' . $this->session->getUuid();
return $this->redirect($redirectUrl);
```
becomes
```
$redirectUrl = $token->{'https://purl.imsglobal.org/spec/lti/claim/target_link_uri'};
$response = $this->redirect($redirectUrl);
$response->headers->setCookie($authCookie);
return $response;
```
- In `Api.js` wherever the headers are explicitly set with `'X-AUTH-TOKEN': authToken,`, we can now  use `credentials: 'include' instead.

I realize there is some duplicate code in the Cookie-making process, and that could still be cleaned up if this is the approach we decide to go.


Ishfaq-code (commented):
All the endpoints seem to function properly using the new credentials method to store the session ID. 

Currently the `postFile` endpoint seems to be broken not due to the changes in this PR, but because the fix for it was merged into `dev` after this PR was made.

Otherwise functionally it seems good, although there may be a better way to store the session token in the frontend.

Issue #1089, opened 2025-12-17
UDOIT Admin Page Loading Issues


dmols:
Related to #1080 but hoping to do a deeper dive here. The UDOIT Admin page currently has issues loading the list of courses, presumably because it is all loaded immediately AND loading every single course into memory on the frontend. We do not have the scale for the amount of courses we're looking at, so we need to consider a server-side rendering approach, rather than client-side, and consider pagination to facilitate the process. 

This change will also make it so the user doesn't have to wait a long time to load the very first page they see. To address this as well, it makes sense to have a front page before all the api calls occur, to also limit the amount of unintentional course lookups when someone accidentally wanders to the admin tool.

Issue #1080, opened 2025-11-17
Unable to load UDIOT report in Canvas


emilypector:
UDOIT report is not loading when scanning a Canvas course. Users have tried scanning only single items, scanning only a few items at a time, scanning only for errors, and closing and reopening the web browser. There are two error messages that display after pressing the "Scan This Course" buton:

1. Please select what you would like to see on your report (even though the "Scan for" items have already been selected).
2. UDOIT failed to scan this course.

PR #1130, opened 2026-03-13
Admin Loading Issues Fix Pt.2

### LINKED ISSUE #1153: Courses Titled "New Course #{lms_course_id}" are Added as Rows in the Course Table
At the moment, there seems to be a bug where rows are being added to the `Course` table titled "New Course #{lms_course_id}", and those courses always seem to have an `lms_course_id` that is a duplicate of a row that exists in the database, and is properly labelled with the Canvas course name. This is currently a minimal data integrity issue and seems to have no negative effects on actual functionality, but in a situation where we want the `lms_course_id` to be a unique column (such as in #1130), then this issue would have to be remediated.

taheralfayad:
Fixes #1153, #1089, #1080

This PR seeks to address the issues with loading courses on the admin panel.

It modifies the courses table by adding the following fields:

- `course_professors`: A JSON field to store the names of professors that teach the course

and modifies pre-existing data by:

- Adding a unique constraint to the composite of `lms_course_id` and `institution_id` for courses
- Updating the columns of `lms_account_id` and `lms_term_id` to become foreign keys to the `Account` and `Term` tables respectively.

Steps to perform migration:

1.) Provided that you have already run the `Version20260304160904` migration, ensure that you have already populated the Account and Term tables. You can populate the `Account` and `Term` tables using the make command `make admin-panel-retrieve-data TABLES="accounts terms"`.

2.) Once you've added the Account and Term tables, you're now ready to run the migration in this pull request. Start off by identifying any `lms_course_id` duplicates in your table. This migration WILL NOT work if you have courses with duplicate `lms_course_id`s. After you've deleted those courses from the database, go ahead and run the migration `Version2026031220211` using `make migrate-to VERSION={2026031220214}`. The migration should run smoothly from there. 

3.) Activate the ETL task that populates the courses table by running `make admin-panel-retrieve-data TABLES="courses"`, and you should now have in your database all the data necessary to power the admin panel.


taheralfayad:
@dmols @mbusch3 I believe this PR is ready for an initial review. There are still features (such as the View report for all courses and the Scan course content) in the admin panel that require some work, but this should at least get the admin panel in a 90% working state.

assets/js/Components/Admin/AdminFilters.js
```diff
@@ -74,57 +71,38 @@ export default function AdminFilters({
...
+            <label htmlFor="inputTerm" className="me-2">
+              {t("filter.label.term")}
+            </label>
           </div>
           <div className="filter-group">
             <select
               id="inputTerm"
               disabled={loadingContent}
-              value={filters.accountId.toString()}
+              value={filters.termId.toString()}
```
Ishfaq-code:
```suggestion
              value={filters?.termId?.toString()}
```

Ishfaq-code (commented):
The setup seems to be working fine of going into admin. Some rendering issues when variables are undefined on the frontend which I commented.

Only issue I faced when I scanned one of the courses it did appear to catch that but the stats do not show up on the courses table

dmols (approved):
Changes look good! Admin panel works as expected. Good work!

Ishfaq-code (approved):
LGTM

mbusch3:
Still to-do:

- Update migration instructions
- Remove the long list of professors from the report when it's a "View Report for All Courses". It's unhelpful.
- Add the account and term filters to the main dashboard... That dashboard reflects data only from the selected term and account, so it makes sense to show it there, too.
- Update the columns on the course table to the following:
  - Course Name, Instructors, [remove account name... it's in the filter above], Last Scanned, Known Barriers, Potential Barriers, [remove Suggestions], Total Files, Barriers Resolved, and Files Reviewed.

There's more after that, but mostly just prettification of the dashboard, and can be part of a separate PR.

taheralfayad:
List of completed to-dos:

- Remove the long list of professors from the report when it's a "View Report for All Courses". It's unhelpful.
- Add the account and term filters to the main dashboard... That dashboard reflects data only from the selected term and account, so it makes sense to show it there, too.
- Update the columns on the course table to the following:
     - Course Name, Instructors, [remove account name... it's in the filter above], Last Scanned, Known Barriers, Potential Barriers, [remove Suggestions], Total Files, Barriers Resolved, and Files Reviewed.

Makefile
```diff
@@ -4,35 +4,107 @@ ifneq (,$(wildcard ./.ins.env))
...
 endif
 
-# spin up the containers
-start:
-	docker compose -f docker-compose.nginx.yml up
+# ──────────────────────────────────────────────
+# Variables
+# ──────────────────────────────────────────────
 
+COMPOSE := docker compose -f docker-compose.nginx.yml
```
mbusch3, line 11: SO much cleaner. Thank you for this. :)

assets/js/Components/Admin/AdminDashboard.js
```diff
@@ -63,188 +59,241 @@ export default function AdminDashboard({
...
     }
 
-    setDashboardStats(stats)
-  }
+    setDashboardStats(stats);
+  };
 
   if (dashboardStats.loading) {
-    return <div className="p-3">Loading dashboard...</div>
+    return <div className="p-3">Loading dashboard...</div>;
```
mbusch3, line 141: This kind of stuff should be sent through translation, like `{t('admin.loading_dashboard')}`

HOWEVER, all of the other text below shouldn't be translated YET until we finalize what is actually going to be on the dashboard.

assets/js/Components/Admin/CoursesPage.js
```diff
@@ -14,349 +14,444 @@ export default function CoursePage({
...
-              textalign="center"
-              className={`btn btn-text btn-icon-only ${((course.loading || isAnyScanning) || !hasReport) ? 'btn-disabled' : ''}`}
-              disabled={(course.loading || isAnyScanning) || !hasReport}
-              title={hasReport ? t('report.button.view_report') : t('report.button.no_report')}
-              aria-label={hasReport ? t('report.button.view_report') : t('report.button.no_report')}
+          accountName: course.accountName || "---",
+          lastUpdated: course.lastUpdated || "---",
+          barriers: hasReport && course.latestReport ? barriers : "---",
+          suggestions: hasReport && course.latestReport ? suggestions : "---",
+          contentFixed:
```
mbusch3, line 119: I don't want to have two separate fields here for "Fixed" and "Resolved." Once a human has looked at it and said they're done, it's done. Add the two together and call it "Barriers Resolved".

assets/js/Components/Reports/ReportsTable.js
```diff
@@ -1,27 +1,39 @@
...
   const headers = [
-    { id: "created", text: t('report.header.date') },
-    { id: "knownBarriers", text: t('report.header.issues'), alignText: 'center' },
-    { id: "potentialBarriers", text: t('report.header.potential'), alignText: 'center' },
-    { id: "filesUnreviewed", text: t('report.header.files_unreviewed'), alignText: 'center', divider: true },
-    { id: "contentHandled", text: t('report.header.items_handled'), alignText: 'center' },
-    { id: "filesReviewed", text: t('report.header.files_reviewed'), alignText: 'center'}
-  ]
-
-  if (isAdmin) {
```
mbusch3, line 19: Is there a reason you dropped the last column? The idea was that when I show this table, there would be a count of the number of courses that have this error to show how widespread it is.

If that data is not easily available, that's fine. But I just wanted to make sure that this wasn't lost on accident.

taheralfayad: The reason that I removed this was because, in the original code, we were passing in `isAdmin={true}` in the `Admin/ReportsPage.js` component, regardless of whether or not the user was viewing a report for a singular course or the report that contains all the courses. Meaning that when you were viewing the reports page for a single course, that column would just contain the value 1, which did not seem that useful in my eyes. That said, I did not account for the fact that this is also reused in the View All Reports page, and the column makes much more sense in that context than when viewing the reports for a singular course. I can restore this column for the View All Reports functionality if that is desired.

assets/js/Components/Admin/AdminDashboard.js
taheralfayad: I agree. I'll push a fix for that as well as adding that translation in `en.json` and `es.json`.

taheralfayad:
Regarding this:

- Update migration instructions

@mbusch3 and I have had discussions regarding how best to roll these changes out in production for UCF, as well as for the open-source project. At the moment, the main hindrance to making this update fully automatic is that there might be rows in the `Course` table that contain duplicate `lms_course_id` and `insitution_id` pairs, which violates the joint uniqueness constraint that I added on `lms_course_id` and `institution_id` pairs. This is due to an unresolved bug where courses titled `New Course #` are added to the database on LTI launch. I can add a line in the migration to automatically delete rows in the Course table that contain the string "New Course", but I personally favored manual remediation just to ensure minimal data loss. I would also recommend that we address the `New Course #` issue prior to merging these changes. Any thoughts?

Ishfaq-code (commented):
Functionally seems good. Some changes that may be good:

1. On the courses table the instructors appear with their login names but after scanning the course it gets changed with their actual first and last name. Maybe confusing to users.

2. Adding a loader to the course being scanned maybe a helpful indicator

3. Add a message or take user to report page of the scanned course after a course has scanned for better indication.

PR #1131, opened 2026-03-13
Adds asynchronous batch scanning

mbusch3:
# THIS IS A DRAFT - DON'T USE UNTIL SCANNER CAN MASSIVELY SCALE

This change significantly improves scanning speed WHEN your local Docker Desktop can handle it. If the settings are too low, it actually INCREASES the amount of request timeout errors that the `udoit3-ace` container returns. Those errors are almost certainly from scan requests that can't get through to the server because the container is overloaded.

### DIFF
--- a/src/Controller/SyncController.php
+++ b/src/Controller/SyncController.php
@@ -147,11 +147,7 @@ class SyncController extends ApiController
 
             // Add rescanned Issues to database
             foreach ($report->getIssues() as $issue) {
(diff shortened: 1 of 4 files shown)


mbusch3:
The biggest issue with serious asynchronous requests is that the Canvas API has a major throttle that makes it trickle out data at a slow and steady stream. Asking for 10 files at a time does not improve the performance.

PR #1134, opened 2026-03-17
Quickfix file state change

Ishfaq-code:
If clicked on "Keep Current File" the first thing in the file form, the state never changed. Fixed it so when clicking a different option changes state properly.


dmols (approved):
Worked perfectly on my end! Pushing to `dev`.

PR #1137, opened 2026-03-24
fix: fixed css so modal doesn't grow with viewport

Ishfaq-code:
Modal was expanding all the way when opened from faculty tool. Fixed it.

### DIFF
--- a/assets/css/udoit4-theme.css
+++ b/assets/css/udoit4-theme.css
@@ -549,12 +549,13 @@
     box-shadow: 0 0 10px var(--shadow-color);
 
     &.dialog-full-screen {
-      width: 95vw;
-      max-width: 95vw;
-      height: 95vh;
-      max-height: 95vh;
-      position: fixed;
-      top: 2vh;
+      width: 95%;
+      max-width: 95%;
+      height: 95%;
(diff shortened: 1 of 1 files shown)


mbusch3:
@Ishfaq-code The issue isn't _just_ that the modal is too tall. The issue is that the entire app is too tall.

The problem comes from Faculty Tools not having fixed sizing; it just makes the `<iframe>` as tall as the content inside. On my machine, Faculty Tools is 6357 pixels tall. And then when you open UDIOT from inside Faculty Tools, it inherits the too-tall `<iframe>`. Most of the time, using `vw` and `vh` units instead of pixels is awesome, but inside an iframe, the "visual" width and height no longer reflect the ACTUAL screen, but the `<iframe>`'s dimensions.

Obviously, your changes fix the modal, but I'd love to actually figure out a way to fix the `<iframe>`, too. Right now we have a `max-height: 1024px` attribute on the `#app-container` in `udoit4-theme.css`, but that's an arbitrary height and has nothing to do with a users's display.

Ishfaq-code:
Initial changes should be fix by `form-ui-changes` branch. Iframe issue might be a issue on it's own

PR #1141, opened 2026-03-26
Remove equal-access-server folder and use equal-access-server image from GHCR in docker-compose-nginx.yml file

panbed:
The `equal-access-server` folder has been in the codebase for a while now, but we've moved development of the server to a new repository, linked here: https://github.com/ucfopen/equal-access-server. We also build images that are published to the GitHub Container Registry.

The `equal-access-server` is not necessarily a UDOIT-only project, and work should be done on the repo instead. Since `docker-compose-nginx.yml` was pointing to the `equal-access-server` folder, it will now build the image from the repository:
```
  ace:
    container_name: udoit3-ace
    image: ghcr.io/ucfopen/equal-access-server:main
    ports:
      - "3000:3000"
    environment:
      - PORT=3000
```

Moving `equal-access-server` into another repository lets us move a lot of the dependabot issues and security vulnerabilities to be part of the scanner rather than UDOIT itself.


dmols (approved):
UDOIT works as expected despite removing the equal access folder, so everything looks good! Only concern is that I still got the 'UserSession error being null' issue, which was initially breaking the app, but I might just add that to this PR.

PR #1143, opened 2026-03-27
Adding to and reworking README.md, INSTALL.md, and INSTALL_CANVAS.md

dmols:
Made changes to the installation instructions as well as the general README, so that it's a little easier to follow along. 

Biggest change is that the scopes are ordered by how they appear on Canvas. That should facilitate setting up the scopes, which tends to be one of the most annoying parts of the API Key setup.


dmols:
My biggest concern here is the images I added. I made them a set size so they wouldn't appear too big on the screen, but I believe this could be inaccessible for other screen sizes. If there's a better way to handle this, I'm open to feedback!

Ishfaq-code (approved):
LGTM

panbed (approved):
I think it looks good, the wording sounds a lot better for the changes you made and seems friendlier to me.

PR #1148, opened 2026-03-31
Adds eslint and prettier

### LINKED ISSUE #1119: Add linter for JavaScript
Make code look prettier

ewainberg:
Set up to ignore prop-types rule (context: https://stackoverflow.com/questions/38684925/react-eslint-error-missing-in-props-validation)
You can run these packages from the terminal. Ideally, you should set up the extensions for each in your preferred code editor. VSCode has them, not sure about others.

Use package.json scripts to run each, for example:
npm run lint -- assets/js/Components/Forms/FormReviewOnly.js
npm run pretty -- assets/js/Components/Forms/FormReviewOnly.js


dmols (approved):
Changes look good! Scripts run as intended. I added a `--cache` flag to each so that each subsequent run of it is a lot quicker, as it will only lint any files that have changed since last time the script was ran. The first time will always take the longest, since it will scan all files.

Also worth noting that `prettier` does come with some implicit standards to follow, but we can add extra rules in a `.prettierrc` file in the future, if we want `prettier` to verify other styling practices.

PR #1149, opened 2026-03-31
Scan Single Courses from Course Report Page

Ishfaq-code:
### Summary
- Scan single course from the course report page. This wasn't scanning because it needed to check if the user is an admin before getting permission to scan
- Display statistics of latest report on course report page tables


taheralfayad (approved):
LGTM 👍

PR #1151, opened 2026-04-07
Finally adds list form

### LINKED ISSUE #1121: Add List Form support
The plan for this is to put a tiny mce text editor on the left side of the screen, scroll focus to the list content, and let the user make their edits.
If this turns out well, a possible addition would be looking for possible lists and pointing them out to the user, and only having that content in the editor instead of the whole page.

ewainberg:
@mbusch3 I was able to fix the saving issue, turned out to be a race condition with processNewReport


ewainberg:
Duplicate

PR #1154, opened 2026-04-10
Dark mode complete, file review page done

mbusch3:
Trying to merge the last of the Dark Mode changes into the `form-ui-changes` branch.


dmols (approved):
Reviewed individually and on a call! I like the changes, and the forms work as we expect. Moving it to `form-ui-changes`.

PR #1159, opened 2026-04-16
List form integration

mbusch3:
The List form is now fully integrated into the new theme and workflow. Major considerations:

1) During the `Report.js` processing, there is now a `Lists.js` service that groups List issues. So instead of three consecutive paragraphs being 3 issues, they are one issue that points to three separate paths for display purposes.
2) Every time any changes are made, they are immediately applied to the `ActiveContentItem`.
3) The extra wrapping and error class for the live preview are applied in the `HtmlPreview.js` file.


Ishfaq-code:
Closing this as this is addressed by #1166

PR #1162, opened 2026-04-21
Split up settings object

### LINKED ISSUE #1161: Break up `settings` object
Currently, the `settings` object handles way too much, including defaults, mutable user preferences, API URLs, current course, and user info. This can become a problem when trying to reuse components that need some of these in a context that does not have access to the rest (e.g., an admin settings page that needs access to user preferences but has no notion of a course and therefore cannot have the same kind of settings object). 

Without splitting it up, different but similar settings objects could still be passed, and a comment could be made on components that need it warning against using certain fields, but this is just hiding the problem.

cj-young:
This is to resolve #1161.


src/Services/InitialStateService.php
```diff
@@ -0,0 +1,77 @@
...
+        $lang = !empty($metadata['lang']) ? $metadata['lang'] : $lang;
+        $lang = array_key_exists('lang', $roles) ? $roles['lang'] : $lang;
+
+        return [
+            'textSpacing'       => $roles['text_spacing'] ?? null,
+            'fontSize'          => $roles['font_size'] ?? null,
+            'fontFamily'        => $roles['font_family'] ?? null,
+            'darkMode'          => $roles['dark_mode'] ?? null,
+            'alertTimeout'      => $roles['alert_timeout'] ?? null,
+            'dailyGoal'         => $roles['daily_goal'] ?? null,
```
mbusch3: I'm not sure where you found the code for `dailyGoal`, `showFilters`, and `viewOnlyPublished`, but they are no longer used.

It doesn't hurt anything to be added here, but we may end up removing them again. Currently, the every preference that can be changed is on the Settings page.

src/Controller/UserController.php
```diff
@@ -42,32 +42,48 @@ public function update(SessionService $sessionService, User $user, Request $requ
...
 
             
             $oldRoles = $user->getRoles();
             $oldLang = $oldRoles['lang'] ?? 'en';
             
-            $newRoles = $userVals['roles'] ?? [];
-            $newLang = $newRoles['lang'] ?? 'en';
 
-            $changeLang = ($oldLang !== $newLang);
+            // Preference IDs on the client differ from DB column names, so we must translate
```
mbusch3: I'm confused as to why this is necessary. There are no columns in the Database for these preferences. They're all just stored in a single column as JSON.

src/Controller/DashboardController.php
```diff
@@ -105,45 +107,21 @@ public function settingsApi(
             $course = $this->createCourse($user->getInstitution(), $lmsCourseId);
         }
 
+        $preferences = $initialStateService->getPreferences($user);
+
         return new JsonResponse([
-            'settings' => $this->getSettings($course),
-            'messages' => $this->util->getUnreadMessages(true),
+            'messages'     => $this->util->getUnreadMessages(true),
```
mbusch3, line 110: I like this breakdown of messages, preferences, labels (translation text), instanceInfo, and formOptions.

I would like to change one thing, though. Now that I'm testing out UDOIT on two different servers (devhub and ucf.test), it would make more sense if the default foreground and background colors were institution-specific. Otherwise, if we set it in the ENV file, every institution must use the same colors, which their specific themes may not have.

Please remove the $_ENV options for colors, and add it to the instanceInfo at the institution level.

src/Services/InitialStateService.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        $institution = $user->getInstitution();
+        $metadata = $institution->getMetadata();
+
+        return [
+            'apiUrl'           => !empty($_ENV['BASE_URL']) ? $_ENV['BASE_URL'] : false,
+            'course'           => $course,
+            'institution'      => $institution,
+            'versionNumber'    => $_ENV['VERSION_NUMBER'] ?? '',
+            'excludedRuleIds'  => !empty($metadata['excludedRuleIds'])
```
mbusch3: We currently do not support excluded rules. You can remove this part.

ALSO, I added a note above about adding the default foreground and background colors to the institution metadata instead of the ENV variable. That will allow each institution to set their own defaults based on individual branding and custom themes.

assets/js/Components/Forms/AltText.js
```diff
@@ -3,10 +3,10 @@ import RadioSelector from '../Widgets/RadioSelector'
 import OptionFeedback from '../Widgets/OptionFeedback'
 import * as Html from '../../Services/Html'
 import * as Text from '../../Services/Text'
+import { UFIXIT_OPTIONS } from '../../Services/Constants'
 
 export default function AltText ({
   t,
```
mbusch3, line 9: I really like how much cleaner the code is when we add specific constants as needed through `import` statements instead of passing them down through multiple layers of components.

While we're at it... Would you please do the same to the `{t}` translation function? I think it should live in either the `Text.js` or `Lang.js` service instead of being declared in `App.js` and passed down in every component across the whole application. I think that would go a long way toward making the code cleaner and more maintainable.

mbusch3 (changes requested):
VERY thorough work. I will continue to review and make further suggestions.

src/Controller/UserController.php
mbusch3: I take this comment back... half way. I was reviewing again this morning and made it down to the migration where you added these columns to the User.

Generally speaking, you've done exactly what ought to be done with relational databases. They work best when every value is stored in a separate column that is designed for its specific type of data. However, I'm going to push back on this migration.

There are two reasons: first is that this is unsorted data. We will never need to run a query like "Select all users where dark_mode = true" or something. It's only needed once you know who your user is, and you're getting additional information about them. Second, we've changed and added preferences many times throughout development, and I expect we'll do it again. I'd rather have a flexible JSON data structure than need to adjust columns all the time.

assets/js/Components/Admin/AdminDashboard.js
```diff
@@ -25,7 +25,7 @@ export default function AdminDashboard({
   })
 
   const progressMeterRadius = () => {
-    switch (settings?.user?.roles?.font_size) {
+    switch (preferences.fontSize) {
```
Ishfaq-code, line 24:
```suggestion
    switch (preferences?.fontSize) {
```
Safer just in case `preferences` becomes `null` or `undefined` at some point. Prevents from crashing

assets/js/Components/Admin/AdminFilters.js
```diff
@@ -16,26 +18,31 @@ export default function AdminFilters({
...
   useEffect(() => {
-    let tempAccountOptions = []
-    for (const acct of Object.values(settings.accounts)) {
-      tempAccountOptions.push({
-        id: acct.id,
-        name: acct.name
-      })
+    if (accounts) {
+      let tempAccountOptions = []
+      for (const acct of Object.values(accounts)) {
```
Ishfaq-code:
```suggestion
      const accountsArr = Object.values(accounts)
      for (const acct of accountsArr) {
```
Cleaner to turn into an array and iterate over that rather than turning it into an array every iteration.

assets/js/Components/Admin/UsersPage.js
```diff
@@ -46,20 +46,20 @@ export default function UsersPage({
...
+    // api.updateUser(user)
+    //   .then((responseStr) => responseStr.json())
+    //   .then((response) => {
+    //     let tempUsers = Object.assign({}, users)
+    //     if (response && response.id) {
+    //       const ind = users.findIndex((el) => { el.id === response.id })
+    //       tempUsers[ind] = response
+    //       setUsers(tempUsers)
+    //     }
+    //   })
```
Ishfaq-code: Remove comments if implementation works as intended.

assets/js/Components/App.js
```diff
@@ -88,29 +80,23 @@ export default function App(initialData) {
...
-          if(data?.labels?.lang) {
-            let newLanguageSettings = Object.assign({}, newSettings)
-            newLanguageSettings.lang = data?.language || newSettings.lang
-            newLanguageSettings.labels = data.labels
-            setSettings(newLanguageSettings)
+          instanceInfo.user = data.user
+          if(data?.labels) {
+            setLabels(data.labels)
           }
           // setUntranslatedMessage({ message: 'msg.settings.updated', severity: 'success', visible: true })
```
Ishfaq-code, line 97:
```suggestion
```

Ishfaq-code (commented):
Works functionally and does intended behavior. Database is updated properly with frontend changes and settings persist on reload. Code refactoring looks great and makes the codebase much cleaner. 

Some stylistic changes (such as removing unused comments) but otherwise looks good!

assets/js/Components/Admin/AdminDashboard.js
Ishfaq-code: Leaving as is for now as we do not want the user to access app with empty preferences. Should make better error handling for better messaging to the user.

mbusch3 (approved):
The only things I ran across in testing involved a few issues with the API for the Admin Pages, which weren't really from this PR and just showed that my database wasn't configured properly. No issues with the pertinent things. Going to approve and merge!

PR #1163, opened 2026-04-22
Adds caption editor for missing media caption issues

ewainberg:


mbusch3:
This has been absorbed into PR #1213.

PR #1174, opened 2026-05-05
Quickfix: PHP 8.5 image builds properly

Ishfaq-code:
Due to the change from php 8.4 to 8.5, the `Dockerfile.build` was not building successfully. Now build correctly after removing `opcache` installation as it comes in php 8.5 binary.


dmols (approved):
Looked at it on Ishfaq's end, will be testing further in `dev` when we deploy to QA.

PR #1178, opened 2026-05-06
Branch merge between form-ui-changes and dev

Ishfaq-code:
Merges dev into form-ui changes. To test it, make sure scan works properly and forms are operational (especially `ContrastForm.js`)


dmols (approved):
Reviewed this on my end and it works as expected. The only thing is that the text spacing option in the settings page does not work, but that can be reviewed prior to merging to `dev`.

PR #1180, opened 2026-05-07
Adding Symfony Mime Back

Ishfaq-code:
Updating versions removed a needed dependency called `Symfony/Mime` which is needed for file replacement. Added that back.


mbusch3 (approved):
I feel great about adding this back in.

PR #1185, opened 2026-05-11
Fixed display data for the ReportsPage.js

mbusch3:
Fixes multiple issues with the Reports page:

- Previously, the chart and table that show barriers over time only showed the TOTAL count of issues, so when issues got fixed, the numbers didn't go down.
- Also fixed an issue where the "Accessibility Barriers Remaining" table had an inaccurate number for Barriers Resolved. The number was from the initial scan, not updated after the `Report.js`'s `analyzeReport` function.


dmols (approved):
Change looks good; the issue count is reflected appropriately now in the reports screen. Not sure if it's already in the target branch or not, but worth noting this branch also includes the fullscreen UDOIT feature. Will merge after getting an additional review.

mbusch3:
This ended up a handling more than intended. Ishfaq's file page updates were merged to the relevant branch, so I pulled them in. After doing that, we caught some related API issues with the Agent-Header. So those got fixed, too.

PR #1194, opened 2026-05-28
Implement standard LTI tables

### LINKED ISSUE #1193: Add standard LTI table implementation and allow multiple LMS instances

cj-young:
Resolves #1193.

This implements the recommended registration, keyset, key, and deployment LTI tables. The keysets, keys, and deployments are currently practically unused but were added to match standard implementations.

The `JWK_BASE_URL` variable has been completely removed. UDOIT works from any instance without requiring any environment variable changes.

A command has been created to initialize institutions/registrations and is documented in the READMEs.

The only non-standard parts of the registration table is the API client ID, secret, and OAuth login endpoint. Typical LTI authentication uses the `grant_type` field of value `client_credentials` to bypass the user consent screen. However, this form of authentication doesn't allow a sufficient level of authorization required by UDOIT (like updating course content).

The registration table and the institution table currently have a one-to-one relationship. Going forward, the institution table should be used primarily for UDOIT-specific data, and the registration table for LMS-interfacing data.


build/nginx/Dockerfile.php.pdo.mysql
mbusch3, line 1: Confession: I tried to run your PR without a full rebuild and kept running into an issue where \SIGINT wasn't defined because pcntl wasn't enabled. Thanks for adding this explicitly to the Docker container. When it's booted up properly, it works perfectly. :)

src/Controller/LtiController.php
```diff
@@ -52,9 +97,84 @@ public function ltiAuthorize(
 
         $this->saveRequestToSession();
 
-        return $this->redirect($this->getLtiAuthResponseUrl());
+        $postParams = $request->request->all();
+        $getParams = $request->query->all();
+        $allParams = array_merge($postParams, $getParams);
+        $iss = $allParams['iss'];
```
mbusch3: Is there a reason for this setup? For instance, are the same parameters handled via POST sometimes and GET other times? I like that this will catch both, but don't know if I'd reverse the order.

According to [https://www.php.net/manual/en/function.array-merge.php](https://www.php.net/manual/en/function.array-merge.php), "If the input arrays have the same string keys, then the later value for that key will overwrite the previous one." I'd like to think that in case of conflict, if someone edits the GET parameters, it can't overwrite what the system puts in the POST.

src/Entity/Institution.php
```diff
@@ -67,36 +64,6 @@ public function __construct()
     }
```
mbusch3, line 66: I like that the Encryption process is handled elsewhere in its own file. Thank you. :)

src/Services/Encryption/SodiumEncryptionService.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace App\Services\Encryption;
+
+class SodiumEncryptionService implements EncryptionServiceInterface
+{
+
+  private string $encodedKey = 'niLb/WbAODNi7E4ccHHa/pPU3Bd9h6z1NXmjA981D4o=';
```
mbusch3, line 8: RED ALERT!!!! Big no-no. This absolutely should be in an environment variable loaded from there.

mbusch3 (changes requested):

src/Controller/LtiController.php
cj-young: In the LTI spec, the data in the initial launch can be sent by either POST or GET. This method was used previously, it has just been moved here. I can't imagine the order mattering, if the LMS follows spec it shouldn't have information in both (or at the very least conflicting information in each). I will edit it to make it more explicitly use one or the other, but in practice it probably won't matter.

src/Services/Encryption/SodiumEncryptionService.php
cj-young: The key was already hard-coded in `User.php`. This is addressed in PR #1078. Should we fix it in this PR or just merge the other one? At least for now, it shouldn't be a huge issue unless someone gets access to our database.

PR #1200, opened 2026-06-09
User session fix

cj-young:
Fix issue with user sessions, allowing users to access other users' preferences.

A new session table was created to handle short-lived persistent LTI launch logic. 

When an LTI session is successfully consumed, it is deleted (and therefore invalidated). However, because errors or early exits can possibly occur after the session is created but before it is consumed, a cron job has been set up to clean up any other expired sessions. The timing of this job (every 2 hours) is arbitrary and can be changed if necessary.


cj-young:
Pulled into wrong branch, and relies on another PR that isn't merged yet. Will open a new one once the LTI table PR is merged.

PR #1202, opened 2026-06-11
Fix file size too large error and refactor API toast messaging

### LINKED ISSUE #1201: Fix error messaging for large files in file form
When you submit a file that is over 10MB in the file form, the server rejects the file (nginx has default upload size limit of 10MB). However, nothing is displayed to the user notifying them that the upload failed.

cj-young:
Fixes #1201.

Previously, the logic to display a toast notification was prop-drilled into each component that makes an API call, and a new API instance is created each time. Now, a single API object is used. Also, the logic for displaying a toast has been moved into the App component in the form of a new event listener on the API object. 

A 413 option for the toast display function was added to add feedback on "File size too large" errors. Other error status toasts can easily be added later.

In the future, error display shouldn't rely just on status codes since very different errors might have the same status codes but completely different contexts and therefore different intended error messages. A more robust error handling method for API calls should be added. This includes abstracting away response handling (e.g. removing the middleman of `response.json()` and checking `response.ok` on every call) and replacing it with API methods that either return the correct result or throw an error.


mbusch3 (approved):
I finished the full review and testing. There was a minor issue due to merging and another where in one case, the api endpoint was using the old `api = new Api` instead of importing the new pre-made `api` object. All found issues have been fixed.

PR #1213, opened 2026-07-06
Add captions flow into File Review page.

mbusch3:
# Caption Editing widget

File Review now includes CAPTION EDITING for media files (audio and video)!

image

## Features

- Existing caption information is loaded from and saved back to Canvas.
- Each file can have multiple tracks associated with it (like "English Captions", "English Descriptive Audio", and "Spanish Captions"). Checks are in place to make sure that no two tracks have the same Language + Type combination.
- Cues can be added at the current timestamp or directly before/after existing captions.
- Live preview of the video includes the current captions/subtitles.
- The waveform visualizer shows current text and accurate timestamps.
- Caption timing can be edited by drag-and-dropping inside the waveform visualizer.
- Drag-and-drop timing also includes "snapping" to align exactly with nearby captions when within 0.25 seconds.
- Media files have an accurate progress bar when downloading.
- Media files are cached so long as the user stays on the File Review page. Closing the editing widget or going to the "next file" doesn't immediately dump the file in case the user wants to go back.
- All controls are keyboard-accessible.

## Testing

This PR can only be tested on the TEST server, since it requires media files be uploaded and processed, and that's where we have that. These media files should show up in the Review Files page. Clicking any `audio` or `video` type file will open the new captioning widget instead of the Review-and-Replace widget.

image


mbusch3:
Based on what Carlos showed in the weekly meeting, I've updated the loading widget in [this commit](https://github.com/ucfopen/UDOIT/pull/1213/commits/86075ebdda6960c8eeaefa2175dca3c91f945ba0).

- If a media file has a "known" file size of < 50K, it is treated as if the file size is unknown. I reproduced Carlos's error, and it seemed like the file size was consistently 12K, but I gave it some wiggle room. If there really are media files of less than 50K, the non-percent loader will do just fine.
- The actual progress bar doesn't have a `width={loadedPercent}%` style attribute, but now a `width=min({loadedPercent}%, 100%)` attribute.
- Also added an additional style to reduce the amount of text jitter when the numbers go up quickly. Doesn't apply correctly to the Sans Serif: Enhanced Readability (Lexend) font, but it's fine on the other three.

assets/js/Components/SettingsPage.js
```diff
@@ -92,7 +92,7 @@ export default function SettingsPage({
 
   return (
   <div className="flex-column settings-page-container h-100">
-    <h1 className="pageTitle pb-2">{t('menu.settings')}</h1>
+    <h1 id="pageTitle" className="pageTitle pb-2">{t('menu.settings')}</h1>
```
carloselopezjr: I think `id="pageTitle"` is unused since `pageTitle` is defined only as a class

assets/js/Services/Captions.js
carloselopezjr, line 1: Image

Out of curiosity, are there plans to support SRT caption files as well? Thought to ask since SRT and VTT are the two most common caption files and Canvas supports both files when uploading media to a page.

assets/js/Components/Captions/MediaCaptionsEditor.js
```diff
@@ -0,0 +1,1233 @@
...
+                      />
+                    </div>
+                    <button
+                      className="btn-secondary btn-small btn-icon-only"
+                      disabled={isDisabled || !cues || cues?.length === 0}
+                      aria-label={t('form.media.button.export_vtt')}
+                      title={t('form.media.button.export_vtt')}
+                      disabled={cues.length === 0}
+                      onClick={handleExport}>
+                      <UploadIcon className="icon-md" />
```
carloselopezjr: Should be `<DownloadIcon/>` here since we're exporting

assets/js/Components/Captions/MediaCaptionsEditor.js
```diff
@@ -0,0 +1,1233 @@
...
+                            title={t('form.media.button.add_cue')}
+                            disabled={isDisabled || error !== ""}
+                          >
+                            <AddIcon aria-hidden="true" className="icon-md" />
+                            {t('form.media.button.enter_with_type', { captionType: captionTypes[vttArray[vttActiveIndex]?.kind] || captionTypes.captions })}
+                          </button>
+                          <button
+                            className="btn-secondary btn-small btn-icon-left flex-shrink-0"
+                            onClick={() => handleImport()}>
+                            <DownloadIcon className="icon-md" />
```
carloselopezjr: Should be `<UploadIcon/>` here since we're importing/uploading

carloselopezjr (changes requested):
- Really like the updated change done to the progress bar when loading a file in response to the feedback given during yesterday's UDOIT meeting. 
- `MediaCaptionsEditor.js` has `<DownloadIcon/>` and `<UploadIcon/>` inverted. They should be swapped with each other 
- Was curious about plans to add support for SRT caption files since canvas natively supports both SRT and VTT.

assets/js/Components/Widgets/FileInformation.js
```diff
@@ -1,37 +1,40 @@
...
-                <div>{t("form.file.label.file_type")}: <span className='fw-bold'>{file.fileType}</span> </div>
-                <div>{t("form.file.label.file_size")}: <span className='fw-bold'>{file.fileSize}</span> </div>
-            </div>
+          :  
+            <div className="fw-bold">{fileName}</div>
+          }
+          <div className='font-smaller flex-row gap-1 pt-1'>
+            <div className='fw-light'>{fileTypeText}</div>
+            <div className='align-items-center'>•</div>
+            <div className='fw-light'>{fileSize}</div>
```
carloselopezjr, line 34: Image

If the media file has the 12KB size error, the incorrect size is displayed on the frontend

carloselopezjr (changes requested):

assets/js/Components/Widgets/FileInformation.js
mbusch3: Fixed file size computation to be based off of actual download size in [e6439b0](https://github.com/ucfopen/UDOIT/pull/1213/commits/e6439b0654ffcce3ab957a3a45097437ac0b8721)

assets/js/Components/Captions/MediaCaptionsEditor.js
mbusch3: Switched upload and download icons in [e6439b0](https://github.com/ucfopen/UDOIT/pull/1213/commits/e6439b0654ffcce3ab957a3a45097437ac0b8721)

assets/js/Components/Captions/MediaCaptionsEditor.js
mbusch3: Switched upload and download icons in [e6439b0](https://github.com/ucfopen/UDOIT/pull/1213/commits/e6439b0654ffcce3ab957a3a45097437ac0b8721)

assets/js/Services/Captions.js
mbusch3: @carloselopezjr This is a GREAT question, that I had to do some research and testing on. Short answer is no, we do not need to support SRT, but not for the reasons that I had suspected.

When users upload an SRT file to Canvas, it automatically converts it to a VTT before it saves it. I've been able to confirm this in Test. So when our tool goes to edit the SRT, it's really just editing a VTT. And it makes no sense to save in SRT format, since it would just be converted anyway.

The good news is that I've confirmed that our tool should work seamlessly regardless of the format of the original upload.

mbusch3: Update: I ran a bunch MORE tests. Turns out there are a few features of SRT that I hadn't accounted for. One of them is pre-formatting multi-line captions. Usually the captions have automatic line breaks when the captions get too long. But if you want line breaks in a specific place, you can do that.

The thing is FINE in the files themselves, but our editor didn't allow for line breaks. So I've updated all of the cue editors from `<input>` elements to `<textarea>` elements. I've verified that the styles look the same, but that when the user hits "Enter" the textbox grows and accommodates multiple lines.

mbusch3:
Testing on 8/21 determined that things are great... in Chrome. But that captions do not show up at all in the preview window on Firefox. I need to see if this is a personal setting somewhere or just a browser-specific issue. Need to do in-depth tests with Safari as well, but I'm having trouble getting the `https://` test server to allow `http://localhost` content.

ewainberg:
I think this is in a good state to be reviewed a final time (for now) to be merged and tested

PR #1226, opened 2026-07-30
Telemetry update

dmols:
Added an option to gather very minimal metrics from the user on app startup. We can change how/when we want to trigger the call in the future, if we think there's a better spot for it. 

The telemetry logic can be seen in [this repo](https://github.com/ucfopen/telemetry-agent). It may get a couple updates in the near future as we add some extra features, but the overlying logic will remain the same. 

Please let me know if you have any questions or concerns!


scripts/start-app.sh
```diff
@@ -0,0 +1,48 @@
...
+
+        case "$choice" in
+             y|Y )
+                TELEMETRY_ENABLED=true
+                echo "TELEMETRY_ENABLED=true" >> "$ENV_FILE"
+                break
+                ;;
+            n|N )
+                TELEMETRY_ENABLED=false
+                echo "TELEMETRY_ENABLED=false" >> "$ENV_FILE"
```
alexanderpaolini, line 26: If you run `make start` with `TELEMETRY_ENABLED=` like the example has, won't we append `TELEMETRY_ENABLED=true|false` to the bottom of the env? Is that fine?

Alternatively, we can use a `grep` to locate the line number, and then `sed` to edit in place. If grep doesn't return a line number, then append as we do before

PR #1230, opened 2026-08-05
Admin Page Rework

Ishfaq-code:
# Summary

PR ready for initial review - expect some UI changes later on.

 - Folder like navigation for accounts for viewing courses and statistics from different accounts
 - Statistics for accounts are calculated on the backend
 - Pagination for `Courses.js`
 - All course report calculation fixed

# Folder navigation for accounts

## Showing Accounts
- Initially the account that loads the admin side is the treated as the 'root' account
- The sub accounts for the 'root' account are loaded 
- Clicking on a sub account loads it's sub account
- New courses and statistics are loaded upon clicking  
- Courses of the selected account AND sub accounts are loaded

## Searching Accounts
- Account search is done through the backend 
- Rendered on frontend
- When clicked after searched, account tree is rendered to show current state of the tree

# Account Statistics Calculated in the Backend
- Upon clicking a new account or selecting a new term stats are recalculated
- Top 5 most inaccessible course and top 5 most recurring 'top barrier' is shown

# Pagination
- Only 10 courses are loaded from the backend at a time
- Pagination information is saved and updated on every backend call
- Searching for courses is also done on the backend and paginated

# Misc
- All terms are loaded onto the frontend dropdown,
- Report schema is updated due to #1236 being merged with this PR
- Account schema updated to account for parent accounts and depth of accounts

# Testing

- Run new migrations for database (accounts and reports) using `make migrate`
- Get data needed for admin page using `make admin-panel-retrieve-data`
- Scan couple of courses on environment where data was collected (subject to automation using panopticon soon)
- Open admin panel on environment where the data was collected
- Interact with data on pages, filtering, etc.

Looking for feedback regarding UI and any bugs or features that may come up or is confusing.


assets/css/udoit4-theme.css
```diff
@@ -913,6 +913,21 @@
     }
   }
 
+  .navigation-breadcrumb {
+    font-size: calc(.8 * var(--font-size));
+    font-weight: 400;
+    margin-inline-start: calc(0.5 * var(--spacer)) !important;
+    margin-inline-end: calc(0.5 * var(--spacer)) !important;
```
mbusch3: These two lines really shouldn't have the `!important` flag. In a quick test on Chrome, removing it didn't change the layout at all.

assets/js/Components/Admin/AdminApp.js
```diff
@@ -90,137 +213,405 @@ export default function AdminApp(initialData) {
...
+
+    return (
+      <div className="admin-account-tree-branch" key={account.lmsAccountId}>
+        <div
+          className={isRoot ? "admin-account-tree-root" : `admin-account-tree-item ${isSelected ? "selected" : ""}`}
+          role={isRoot ? undefined : "button"}
+          style={{ "--account-depth": depth }}
+          tabIndex={isRoot ? undefined : "0"}
+          onClick={isRoot ? undefined : () => handleAccountSelect(account, depth)}
+          onKeyDown={
```
mbusch3: After playing with the navigation, I would really love it if clicking on the root selected the root. For instance, once I start digging through the tree, the primary way to go back to the root account is using the breadcrumbs. You can also re-click on a 1-level-down descendant, but with the current UI, that doesn't like an intuitive way to get back to the top.

assets/js/Components/App.js
```diff
@@ -161,8 +161,6 @@ export default function App(initialData) {
 
     api
       .setReportData(tempReport.id, {
```
mbusch3: This part needs to be revisited.

After the front end performs a scan, the results are amended in the `AnalyzeReport` function. Since we are now storing more precise counts, those need to be added to the `setReportData` API endpoint.

As an example: I just scanned a course that hadn't been scanned before. In that course's [summary page](https://devhub.cdl.ucf.edu/courses/13/external_tools/305), it reports 21 issues and 53 potential issues.

Image

But in the admin course report, it shows 21 issues and *151* potential issues.

Image

The next chart, which shows specific errors, is fine and the data is correct. Just the counts from the `report` table.

assets/js/Components/Widgets/DashboardCourseTable.js
```diff
@@ -0,0 +1,88 @@
...
+        ascending: false,
+        pageNum: 0,
+      })
+    
+    const headers = [
+        { id: "title", text: "Course Name" },
+        { id: "totalActiveIssues", text: "Barriers"},
+        { id: "scanRule", text: "Most Recurring Barrier"},
+        {id: "lastUpdated", text: "Last Updated"}
+    ]
```
mbusch3: Strings need to be in the `en.json` file and not hardcoded. (Same on line 77)

assets/js/Components/Widgets/DashboardCourseTable.js
```diff
@@ -0,0 +1,88 @@
...
+             href="#reports"
+             onClick={(event) => {
+               event.preventDefault()
+               handleReportClick(course)
+             }}
+           >
+             {course.title}
+           </a>
+         ),
+         scanRule: t(`form.${formNameFromRule(course.scanRule)}.title`),
```
mbusch3: This will require custom logic:

Image

I had a course with no errors. I verified that the `course.scanRule` was `""`, so it should be blank instead of "Unable to Fix in UDOIT". But the `formNameFromRule` function is a switch statement with a default. If there is no `course.scanRule`, the "scanRule" text here should be something like "---" to indicate no data.

assets/js/Components/Widgets/DashboardScanRuleTable.js
```diff
@@ -0,0 +1,70 @@
...
+        ascending: true,
+        pageNum: 0,
+    })
+
+    const headers = [
+        { id: "rank", text: "Rank" },
+        { id: "rawRule", text: "Equal Access Scanner Rule" },
+        { id: "normalizedRule", text: "Normalized Rule" },
+        { id: "count", text: "Number of Courses" }
+    ]
```
mbusch3: All text should be in the `en.json` file and not hard-coded. (See also line 59)

Additionally, now that all of the rules will have normalized text, I would prefer to JUST have that column and remove the "rawRule" column.

assets/js/Components/Admin/AdminDashboard.js
```diff
@@ -1,141 +1,34 @@
...
-          if (scanDate > thirtyDaysAgo) {
-            stats.recentScans++;
-          }
+    if(dashboardStats?.scanRanked){
+      const tempRanked = []
+      for(const rule in dashboardStats.scanRanked){
+        if(rule){ 
+          tempRanked.push({
+            rawRule: rule,
+            normalizedRule: t(`form.${formNameFromRule(rule)}.title`),
```
mbusch3: I'd like to remove the "rawRule" column and just have the normalizedRule.

Furthermore, I'd love for the normalizedRule value to be a JSX object that has the text AND an `<InfoPopover>` widget with the rule summary. Feel free to shamelessly copy the code from `ReportsPage.js` starting at line 88.

assets/js/Components/Admin/AdminFilters.js
```diff
@@ -2,116 +2,105 @@ import React, { useState, useEffect } from "react";
...
+  accountStack,
+  handleAccountSelect,
+  setSelectedTerm
 }) {
-  const [accountOptions, setAccountOptions] = useState([]);
-  const [termOptions, setTermOptions] = useState([]);
 
-  // When the "termInfo" and "accounts" are loaded, create the Account and Term dropdown options
+  const [termOptions, setTermOptions] = useState([])
+  const [pendingSearchTerm, setPendingSearchTerm] = useState(searchTerm)
```
mbusch3: In my testing, the whole `pendingSearchTerm` with the `SEARCH_DEBOUNCE_MS` never fired. Since we now have the explicitly called search button (or hitting enter), can we remove the unused code?

Specifically, when I set breakpoints on the page, line 29 was always true, with both variables evaluating to `""`.

assets/js/Components/Admin/AdminHeader.js
```diff
@@ -10,62 +10,42 @@ export default function AdminHeader({
   navigation,
   handleNavigation,
 }) {
+
+
+  const links = [
+    {name: t('Dashboard'), icon: HomeIcon, key: 'dashboard'},
+    {name: t('Courses'), icon: ContentAssignmentIcon, key: 'courses'}
+  ]
```
mbusch3: There's a weird issue where if I am on the dashboard, then I click on a course name, it takes me to that course's report. But then I am no longer on either the Dashboard or the Courses page.

Image

assets/js/Components/Admin/CoursesPage.js
```diff
@@ -6,382 +6,147 @@ import ReportIcon from "../Icons/ReportIcon";
...
-  });
   const headers = [
     {
       id: "courseName",
       text: t("report.header.course_name"),
       alignText: "center",
     },
-    { id: "instructors", text: "Instructors", alignText: "center" },
+    { id: "instructors", text: "Instructors", alignText: "center", sortable: false },
     {
```
mbusch3: Would it be possible to make every column with text or numbers sortable?

assets/js/Components/ReportsPage.js
```diff
@@ -161,7 +161,7 @@ export default function ReportsPage({
       { id: "potentialBarriers", text: t('report.header.potential'), alignText: 'center' },
       { id: "filesUnreviewed", text: t('report.header.suggestions'), alignText: 'center' },
       { id: "contentHandled", text: t('report.header.items_handled'), alignText: 'center' },
-      { id: "filesReviewed", text: t('report.header.files_reviewed'), alignText: 'center'}
+      { id: "reviewedFiles", text: t('report.header.files_reviewed'), alignText: 'center'}
```
mbusch3: My ADD says that if you change this to "reviewedFiles" you should change line 162 to "unreviewedFiles". :)

src/Controller/ReportsController.php
```diff
@@ -116,13 +115,6 @@ public function setReportData(
             }
 
             $data = json_decode($request->getContent(), true);
```
mbusch3: Glad to see the report's 'scanRules' go. Thank you for cleaning this.

src/Controller/SyncController.php
```diff
@@ -30,15 +30,7 @@ class SyncController extends ApiController
...
-        $previousReport = $course->getLatestReport();
-        if ($previousReport) {
-            $data = json_decode($previousReport->getData());
-            $currentVersionNumber = !empty($_ENV['VERSION_NUMBER']) ? $_ENV['VERSION_NUMBER'] : '';
-            if (isset($data->versionNumber) && $data->versionNumber === $currentVersionNumber) {
-                return true;
-            }
-        }
-        return false;
+        return (bool) $course->getLatestReport();
```
mbusch3: This previous logic wasn't to check if the REPORT was the current version, but rather if the APP was the current version. In other words, whether or not the course page data has been changed or the app *thinks* it should rescan, if the version of UDOIT has changed, it forces a full, fresh rescan. This is in case the scan rules, the forms, the database schema, or anything else changes.

src/Repository/CourseRepository.php
```diff
@@ -5,6 +5,7 @@
 use App\Entity\Course;
 use App\Entity\User;
 use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
+use Doctrine\ORM\Tools\Pagination\Paginator;
```
mbusch3: Well, I wish I'd known about THAT!

mbusch3 (commented):
What a HUGE undertaking! I'm really, really proud of how well this seems to be working. 

I tried to put all my comments next to the relevant code, but couldn't attach this one anywhere:

Before this gets merged, I'd like to figure out why the API key from Canvas isn't able to be used directly from the `.ins.env` file with the `AdminPanelRetrievalCommand.php` script.

PR #1246, opened 2026-09-11
Adds titles for all non-form errors, uses them to the Reports screen.

mbusch3:
## UPDATE:

As a part of this rewrite, I did a deep-dive on the IBM checker rules. Since the beginning of this project, some have been deprecated, some changed, and a few added. I've updated the project to reflect that. For instance:

- In the `EqualAccessService.php` file, I've changed the list of rules that get skipped by the scanner. I've added `html_lang_valid`, `page_title_valid`, and `skip_main_described`, which are not applicable to Canvas page content, and removed `aria_accessiblename_exists`, `a_target_warning`, and `aria_landmark_name_unique`, since those can (I hope) be handled in the scanner.

The updated ruleset I used is on the [IBM Accessibility Requirements](https://www.ibm.com/able/requirements/checker-rule-sets/) page, using the 13 July 2026 build and the WCAG 2.1 ruleset.

---

Issue: Any error that doesn't belong to a form is lumped together in the "Unable to Fix in UDOIT" category, which is fine. But on the Reports, they are labelled as "unhandled errors" with their IBM scanner error id:

image

This PR:
- Adds titles for all non-form errors
- Uses the titles on the Reports screen
- Updates the Spanish labels (`es.json`) with corresponding placeholders.

image


mbusch3:
All right, I feel a bit silly, but there is a small subset of issues that still don't have titles. These are issues that are SUPPOSED to belong to a form, but the form has not yet been implemented. In `Services\Ufixit.js`, there is a `rulesToFormName` object, that has a lot of rules that are commented out:

image

So as an example, the `aria_attribute_allowed` and `aria_attribute_required` rules are supposed to be resolved in the `ARIA_ATTRIBUTE` form, but since that form doesn't exist yet, the rules don't have proper titles applied.

